Issue #13, opened 2022-01-12
Add basic support for SSR


davesmith00000:
Should just be a case of providing a method you can call that performs the presentation work based on a given model. Then you could use the same codebase for SPA's or SSR.

davesmith00000:
The complication here is that you'd expect to use this in JVM based services (Http4s for Scala.js does exist but that won't be people's first thought), and Tyrian uses snabbdom to render.

Couple of options:

1. We don't need a virtualdom differ for straight rendering, so we could just write a simple renderer. Probably not hard, but makes it inconsistent with the SPA version.
2. We write a virtualdom + renderer in Scala.js and remove snabbdom.

davesmith00000:
A minimal version is now in the main branch.

PR #16, opened 2022-01-15
Not quite good enough SSR.

davesmith00000:
Work in progress.

This first commit is just a little PoC of how easy rendering html fragments actually is, but by itself it isn't good enough.

Quickest route to market: Tyrian's little SSR solution will need to be published for JVM (maybe JS but maybe not), since you'd want to use it with something like Http4s. This means a multi-project build (Most of Tyrian is not JVM compatible) with cross publishing. At minimum.

We could also do with a nice working example / demo to show how it works.

What this does also mean however, is that Tyrian will have two different rendering systems, one for SSR and one for SPAs via Snabbdom, meaning inconsistent rendering results. The differences should be slight... but then again they might not be!

Maybe that isn't a problem?

If it is a problem, then we may want to do away with Snabbdom and write a pure Scala.js renderer. How hard can it be? 😅


davesmith00000:
Related to issue #13

Issue #18, opened 2022-01-15
Generate common styles


davesmith00000:
It would be nice to generate things like:

`Style("font-family" -> "Arial, Helvetica, sans-serif")`

..to help avoid the typos.

davesmith00000:
A basic version of this is in, you can now do:

```scala
import tyrian.CSS.*

CSS.`font-family`("Arial, Helvetica, sans-serif")
```

Which is the equivalent of the example above:

```scala
Style("font-family" -> "Arial, Helvetica, sans-serif")
```

But is discoverable through your IDE and removes at least one classes of typo. No doubt more will be done in the future.

PR #22, opened 2022-01-17
CSS Property generation

davesmith00000:
Working on #18.

So far basic CSS Properties are being generated. This is already a good start as it removes the possibility of some forms of typo and gives you IDE discovery/completion on the names.

The main question is do we want to do better on the values. Some of the possible values are very complicated.


davesmith00000:
Decided this was a good start, will wait for further requirements to emerge.

PR #27, opened 2022-01-20
fixing reload loop, sbt refactor

Angel-O:
### Summary

- calling the `localPublish` command alias (that was delegating to the sub-projects), was causing the settings to be constantly reapplied, probably because the command was called on the `aggregate` project.
- `build.sbt` refactor


davesmith00000 (approved):
That is a lovely refactor @Angel-O and it works perfectly - many thanks! 😄

Issue #6, opened 2021-06-08
Add WebSocket support


davesmith00000:

davesmith00000:
Done, will be in the next release.

PR #30, opened 2022-01-27
WebSocket Support

davesmith00000:
This is a sort of port of the Indigo solution, Tyrian-ified.. Very much a work in progress.


davesmith00000:
Where I'm up to:

- Not sure I like the integration pattern...
- Connection to the socket server is established via a subscription.
- Messages can be sent to the service via a command.
- Messages are received from the server but are then lost because I'm not feeding them back into the process correctly.

davesmith00000:
This work is towards issue: #6

davesmith00000:
I think this is ready for a first pass at a review. It started off as a port of the Indigo implementation but is now a very different beast.

hobnob (approved):

PR #43, opened 2022-02-02
Fixed #34: Launch mechanism similar to Indigo's

### LINKED ISSUE #34: Launch mechanism
When Indigo starts, instead of using `main`, you embed the game and invoke a launch method. The advantage of this is that you can supply (crude) flags (like command line args) during start up, as well as the ID of the element to hook onto.

Should Tyrian do a similar thing?

davesmith00000:
Rather than just embedding your app as a script, you would now launch it, thus allowing you to pass in the container id and `String->String` flags (like Indigo does - think command line args).

Might be worth noting that aside from being much less clever, this is straight up stolen from Elm. It essentially changes the `init` function from a value `(Model, Cmd[Msg])` to a function `Map[String, String] => (Model, Cmd[Msg])` allowing you to read the flags at initialisation time and set your model accordingly.

### DIFF
--- a/sandbox/src/main/scala/example/Sandbox.scala
+++ b/sandbox/src/main/scala/example/Sandbox.scala
@@ -1,17 +1,17 @@
 package example
 
-import tyrian._
-import tyrian.Html._
-import tyrian.websocket._
-import org.scalajs.dom.document
-import org.scalajs.dom.Element
+import tyrian.Html.*
+import tyrian.*
+import tyrian.cmds.Logger
+import tyrian.websocket.*
 
+import scala.scalajs.js.annotation.*
+
+@JSExportTopLevel("TyrianApp")
 object Sandbox extends TyrianApp[Msg, Model]:
 
-  def container: Element = document.getElementById("myapp")
-
-  def init: (Model, Cmd[Msg]) =
-    (Model.init, Cmd.Empty)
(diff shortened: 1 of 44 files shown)


sandbox/src/main/scala/example/Sandbox.scala
```diff
@@ -1,17 +1,17 @@
...
+import scala.scalajs.js.annotation.*
 
-  def container: Element = document.getElementById("myapp")
+@JSExportTopLevel("TyrianApp")
+object Sandbox extends TyrianApp[Msg, Model]:
 
-  def init: (Model, Cmd[Msg]) =
-    (Model.init, Cmd.Empty)
+  def init(flags: Map[String, String]): (Model, Cmd[Msg]) =
+    (Model.init, Logger.info(flags.toString))
```
davesmith00000, line 14: So here we've accepted flags and logged them out.

sandbox/tyrianapp.js
```diff
@@ -0,0 +1,3 @@
+import {TyrianApp} from './.js/target/scala-3.1.1/sandbox-fastopt.js';
+
+TyrianApp.launch("myapp", {"message": "Hello, Tyrian!"});
```
davesmith00000, line 3: This is how we launch the app, tell it what container id to attach itself too, and pass in some flags.

hobnob (dismissed):

PR #53, opened 2022-02-23
WebSocket: add keep-alive support

### LINKED ISSUE #52: Add KeepAlive to WebSockets
@gvolpe has done a great job of improving the WebSocket implementation in this PR by adding keep alive functionality:

https://github.com/gvolpe/trading/pull/55

We should try and move the changes up to Tyrian itself.

gvolpe:
closes #52 

Besides adding a keep-alive connection mechanism, I made a change so the WebSocket connection has to be initiated explicitly. The side-effect of creating the connection was previously performed when invoking the `subscribe` method. Now there is no way to call `subscribe` without having a WS connection, which can be done via any of the `WebSocket.connect` variants.

The side-effect is still there but now the name `connect` tells you that is going to happen. I don't know if there's a mechanism in Tyrian where this can be managed by a `Task` like `Cmd` or `Sub`, so for now I live it as is and return `Either[String, WebSocket]`.

I wrote the initial implementation in JavaScript: https://github.com/gvolpe/trading/blob/main/web-app/js/WsPorts.js

There are still a few other things that can be improved in this Scala version. ~~For example, when the user closes the browser tab, that is properly handled in the JS version. I need to play around a bit more and see if the Scala version behaves as expected.~~ UPDATE: it works. I added a few improvements in the second commit. 

~~Also, the keep-alive timeout could probably be made configurable as well but I think this is a good first step.~~ UPDATE 2: Added a `KeepAliveSettings` to configure the timeout and a few other things; it simplified things :)

Let me know what you think, and thanks a lot for working on Tyrian! Really good first experience in ScalaJS for me :)

### DIFF
--- a/tyrian/js/src/main/scala/tyrian/websocket/WebSocket.scala
+++ b/tyrian/js/src/main/scala/tyrian/websocket/WebSocket.scala
@@ -7,6 +7,8 @@ import tyrian.Task
 import tyrian.websocket.WebSocketEvent
 import util.Functions
 
+import scala.concurrent.duration.*
+
 final class WebSocket(liveSocket: LiveSocket):
   def disconnect[Msg]: Cmd[Msg] =
     Cmd.SideEffect(() => liveSocket.socket.close(1000, "Graceful shutdown"))
@@ -39,36 +41,36 @@ object WebSocketReadyState:
       case _ => CLOSED
     }
 
+final case class KeepAliveSettings(message: String, timeout: FiniteDuration, enabled: Boolean)
+object KeepAliveSettings:
+  def default  = KeepAliveSettings("{ \"Heartbeat\": {} }", 20.seconds, true)
+  def disabled = default.copy(enabled = false)
+
 object WebSocket:
   /** Acquires a WebSocket connection with default keep-alive message */
   def connect(address: String): Either[String, WebSocket] =
-    newConnection(address, None, None, true).map(WebSocket(_))
+    newConnection(address, None, KeepAliveSettings.default).map(WebSocket(_))
 
   /** Acquires a WebSocket connection with default keep-alive message and a custom message onOpen */
   def connect(address: String, onOpenMessage: String): Either[String, WebSocket] =
-    newConnection(address, Option(onOpenMessage), None, true).map(WebSocket(_))
+    newConnection(address, Option(onOpenMessage), KeepAliveSettings.default).map(WebSocket(_))
 
   /** Acquires a WebSocket connection with custom keep-alive message */
-  def connect(address: String, keepAliveMessage: Option[String]): Either[String, WebSocket] =
-    newConnection(address, None, keepAliveMessage, true).map(WebSocket(_))
+  def connect(address: String, keepAliveSettings: KeepAliveSettings): Either[String, WebSocket] =
+    newConnection(address, None, keepAliveSettings).map(WebSocket(_))
 
   /** Acquires a WebSocket connection with a custom keep-alive message and a custom message onOpen */
-  def connect(address: String, onOpenMessage: String, keepAliveMessage: Option[String]): Either[String, WebSocket] =
-    newConnection(address, Some(onOpenMessage), keepAliveMessage, true).map(WebSocket(_))
-
-  /** Acquires a WebSocket connection with disabled keep-alive mechanism */
-  def noKeepAlive(address: String): Either[String, WebSocket] =
-    newConnection(address, None, None, false).map(WebSocket(_))
+  def connect(address: String, onOpenMessage: String, keepAliveSettings: KeepAliveSettings): Either[String, WebSocket] =
+    newConnection(address, Some(onOpenMessage), keepAliveSettings).map(WebSocket(_))
 
   private def newConnection(
       address: String,
       onOpenSendMessage: Option[String],
-      withKeepAliveMessage: Option[String],
-      keepAliveEnabled: Boolean
+      settings: KeepAliveSettings
   ): Either[String, LiveSocket] =
     try {
       val socket    = new dom.WebSocket(address)
-      val keepAlive = new KeepAlive(socket, withKeepAliveMessage, 20000)
+      val keepAlive = new KeepAlive(socket, settings.message, settings.timeout)
 
       val subs =
         Sub.Batch(
@@ -82,13 +84,13 @@ object WebSocket:
             Some(WebSocketEvent.Error("Web socket connection error"))
           },
           Sub.fromEvent("close", socket) { e =>
-            if keepAliveEnabled then keepAlive.cancel() else ()
+            if settings.enabled then keepAlive.cancel() else ()
             val ev = e.asInstanceOf[dom.CloseEvent]
             Some(WebSocketEvent.Close(ev.code, ev.reason))
           },
           Sub.fromEvent("open", socket) { _ =>
             onOpenSendMessage.foreach(msg => socket.send(msg))
-            if keepAliveEnabled then keepAlive.run() else ()
+            if settings.enabled then keepAlive.run() else ()
             Some(WebSocketEvent.Open)
           }
         )
@@ -99,7 +101,7 @@ object WebSocket:
         Left(s"Error trying to set up websocket: ${e.getMessage}")
     }
 
-  final class KeepAlive(socket: dom.WebSocket, msg: Option[String], timeout: Int):
+  final class KeepAlive(socket: dom.WebSocket, msg: String, timeout: FiniteDuration):
     @SuppressWarnings(Array("scalafix:DisableSyntax.var"))
     private var timerId = 0;
 
@@ -107,8 +109,8 @@ object WebSocket:
     def run(): Unit =
       if socket != null && WebSocketReadyState.fromInt(socket.readyState).isOpen then
         println("[info] Sending heartbeat 💓")
-        socket.send(msg.getOrElse("{ \"Heartbeat\": {} }"))
-      timerId = dom.window.setTimeout(Functions.fun0(() => run()), timeout)
+        socket.send(msg)
+      timerId = dom.window.setTimeout(Functions.fun0(() => run()), timeout.toMillis.toDouble)
 
     def cancel(): Unit =
       if (timerId <= 0) then dom.window.clearTimeout(timerId) else ()
--- a/tyrian/js/src/main/scala/tyrian/websocket/WebSocket.scala
+++ b/tyrian/js/src/main/scala/tyrian/websocket/WebSocket.scala
@@ -4,48 +4,84 @@ import org.scalajs.dom
 import tyrian.Cmd
 import tyrian.Sub
 import tyrian.Task
+import tyrian.websocket.WebSocketEvent
+import util.Functions
 
-import scala.collection.mutable
-
-final class WebSocket(address: String, onOpenSendMessage: Option[String]):
-
-  @SuppressWarnings(Array("scalafix:DisableSyntax.var", "scalafix:DisableSyntax.null"))
-  private var liveSocket: LiveSocket = null
-
+final class WebSocket(liveSocket: LiveSocket):
   def publish[Msg](message: String): Cmd[Msg] =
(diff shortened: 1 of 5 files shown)


davesmith00000:
> I don't know if there's a mechanism in Tyrian where this can be managed by a `Task` like `Cmd` or `Sub`, so for now I live it as is and return Either[String, WebSocket].

Tyrian does indeed have a very simple `Task` implementation, that's been then since the dawn of time:

https://github.com/PurpleKingdomGames/tyrian/blob/main/tyrian/js/src/main/scala/tyrian/Task.scala

And this is basically how `Cmd`s run behind the scenes:

https://github.com/PurpleKingdomGames/tyrian/blob/main/tyrian/js/src/main/scala/tyrian/Cmd.scala#L56

gvolpe:
Sorry, I wasn't clear enough. I am aware of `Task` as the underlying effect for `Cmd` and `Sub`, but I don't think there's a way for suspending the side-effect of creating a connection while retaining the produced value.

These are the two methods on `Task` I can use, besides the one that takes an `Observable`.

```scala
  final case class SideEffect(thunk: () => Unit) extends Task[Nothing, Unit]

  /** A task that succeeded with the given `value` */
  final case class Succeeded[A](value: A) extends Task[Nothing, A]
```

I don't see how I can create a WebSocket connection in `Task` besides using `Succeeded`, which kind of defeats the point.

```scala
  private def newConnection(
      address: String,
      onOpenSendMessage: Option[String],
      settings: KeepAliveSettings
  ): Task[String, LiveSocket] =
    try {
      val socket    = new dom.WebSocket(address)
      val keepAlive = new KeepAlive(socket, settings.message, settings.timeout)

      val subs =
        Sub.Batch(
          Sub.fromEvent("message", socket) { e =>
            Some(WebSocketEvent.Receive(e.asInstanceOf[dom.MessageEvent].data.toString))
          },
          Sub.fromEvent("error", socket) { e =>
            val msg =
              try { e.asInstanceOf[dom.ErrorEvent].message }
              catch { case _: Throwable => "Unknown" }
            Some(WebSocketEvent.Error("Web socket connection error"))
          },
          Sub.fromEvent("close", socket) { e =>
            if settings.enabled then keepAlive.cancel() else ()
            val ev = e.asInstanceOf[dom.CloseEvent]
            Some(WebSocketEvent.Close(ev.code, ev.reason))
          },
          Sub.fromEvent("open", socket) { _ =>
            onOpenSendMessage.foreach(msg => socket.send(msg))
            if settings.enabled then keepAlive.run() else ()
            Some(WebSocketEvent.Open)
          }
        )

      Task.Succeeded(LiveSocket(socket, subs))
    } catch {
      case e: Throwable =>
        Task.Failed(s"Error trying to set up websocket: ${e.getMessage}")
    }
```

Do you think it's possible to add a `Task.delay` or `Task.suspend`? 

---

Also, it might be too early to discuss this, but I was wondering how we could integrate code written using Cats Effect (or other effects library) in Tyrian. Say I have an HTTP Client (I know Tyrian provides one, just an example) and I want to re-use it. I'm just going to leave this here fow now but I wanted to bring it up at some point 😄

davesmith00000:
Not a full answer to your comment (I shall have to come back to it tomorrow), just wanted to mention the Cats-effect question: I'm not against the idea, but I'm undecided.

I too find the current `Task` implementation (which I inherited) quite particular, a bit hard to work with, and not really what you think of as an effect monad.

Then again, I do like that Tyrian is very low dependancy. Needs more thought.

gvolpe:
@davesmith00000 I was not suggesting a dependency on Cats Effect or any other library, I think it's perfect as it is. I was wondering what the integration with Tyrian would look like. In any case, that would at least belong in a different module but I'd even push it to a different repository to have independent releases and avoid maintenance burden.

Anyway, forget about that for now, that's something I intend to bring up at some point when I have more of a formulated idea 😄 

---

The problem at hand is the suspension of the side-effect of creating a WS connection, which can't be done with `tyrian.Task` at the moment. A `Task.Delay` will do - similar to `Task.SideEffect` but retaining the produced value. 

If that's something we want, I can dig and see if I can implement it.

```scala
/** A task that is a fire and forget side effect */
final case class SideEffect(thunk: () => Unit) extends Task[Nothing, Unit]

/** A task that suspends the side-effect of producing either A on success or E on failure */
final case class Delay[E, A](thunk: () => A, onError: Throwable => E) extends Task[E, A]
```

Alternatively, we can fix the error type to either `String` or `Throwable`.

```scala
/** A task that suspends the side-effect of producing A. If it fails, an error message is returned instead */
final case class Delay[A](thunk: () => A) extends Task[String, A]
```

Thoughts?

armanbilge:
Noticed this thread, I hope you don't mind me popping in with my 2 cents.

> I too find the current Task implementation (which I inherited) quite particular, a bit hard to work with, and not really what you think of as an effect monad.
> 
> Then again, I do like that Tyrian is very low dependancy. Needs more thought.

We've put _a lot_ of effort into making `IO` awesome on JavaScript. For one, it was the Cats Effect test suite that picked up the issues with the Scala.js global `ExecutionContext` that led to its deprecation in [Scala.js 1.8.0](https://www.scala-js.org/news/2021/12/10/announcing-scalajs-1.8.0/) and the [macrotask-executor](https://github.com/scala-js/scala-js-macrotask-executor) project.

Besides everything that makes the `IO` monad great, the JS `IORuntime` maintains feature-parity with the JVM runtime. This includes the ability to obtain fiber traces/dumps that make it possible to introspect all the asynchronous execution graphs in your program while it's running. Since effectively everything on JS is async, this is a big win for debugging. See the "Tracing on Scala.js" section in the [CE 3.3.0 release notes](https://github.com/typelevel/cats-effect/releases/tag/v3.3.0).

FWIW, the cats-effect-kernel which defines the core typeclasses is a pretty low dependency itself (its only dependency is cats) and is not tied to `IO` in anyway.

> Say I have an HTTP Client (I know Tyrian provides one, just an example) and I want to re-use it.

Just dropping this here 😉 https://github.com/http4s/http4s-dom

gvolpe:
@davesmith00000 I took the latest proposal for a spin, code changes are quite minimal. I think this could make for a potential first version and we can improve from there.

Cats Effect is very appealing on ScalaJS after all those improvements @armanbilge pointed out, but you probably want to think that thoroughly before making any decision :)

davesmith00000:
I'll give the CE stuff some thought.

In terms of producing the value from a `Task`, the observable version has an `onNext` and `onError` that are used to great the next values. So on success (`onNext`) you can store the connection, it would then be given to a `Msg`, returned to the update function for you to store.

Would that work?

Example:
https://github.com/PurpleKingdomGames/tyrian/blob/main/tyrian/js/src/main/scala/tyrian/cmds/Dom.scala#L27

gvolpe:
@davesmith00000 not very clear to me, you mean creating an observable task first? If so, how can I construct it? Would it suspend the side-effects as well?

davesmith00000:
I'm running out of week I'm afraid and I don't want to send you on wild goose chase. I'll check out your branch and have a play with it in the morning.

davesmith00000:
@gvolpe Your code actually appears to work perfectly, you just need to change this:

```diff
diff --git a/tyrian/js/src/main/scala/tyrian/runtime/TaskRunner.scala b/tyrian/js/src/main/scala/tyrian/runtime/TaskRunner.scala
index 9957be3..20a42fa 100644
--- a/tyrian/js/src/main/scala/tyrian/runtime/TaskRunner.scala
+++ b/tyrian/js/src/main/scala/tyrian/runtime/TaskRunner.scala
@@ -15,7 +15,7 @@ object TaskRunner:
   def execTask[Err, Success](task: Task[Err, Success], notifier: Either[Err, Success] => Unit): Unit =
     task match
       case Task.SideEffect(thunk)         => thunk()
-      case Task.Delay(thunk)              => try { Right(thunk()) } catch { case NonFatal(e) => Left(e.getMessage) }
+      case Task.Delay(thunk)              => notifier(try { Right(thunk()) } catch { case NonFatal(e) => Left(e.getMessage) })
       case Task.Succeeded(value)          => notifier(Right(value))
       case Task.Failed(error)             => notifier(Left(error))
       case Task.RunObservable(observable) => observable.run(asObserver(notifier))
```

davesmith00000:
@gvolpe I'm still thinking about it, but this bit here feels a bit funky:

```
try ... catch { case NonFatal(e) => Left(e.getMessage) }
```

Notionally, Tyrian doesn't handle exceptions. We have no monadic effect type floating around that allows us to `flatMap` over a monad that has captured an error to be dealt with later or anything of that sort. Doesn't fit the Elm-esque API shape.

Thinking about this in terms of a catamorphism, what where saying here really is that if we destroy/run a task in combination with a `Either[Error, Result] => Msg` function, then we're guaranteeing that we always produce a usable value.

In `Cmd`s like `Http`, this 'guarantee' is met by handling the exception in the `Task` body definition - so that's fine - but doesn't work so well on a delayed effect.

Back to catamorphisms: The implication is (perhaps?) that we need to tell the user that they need to provide a default `Msg` to return in the case of catastrophic failure (i.e. an exception) as part of `Task.Delayed`.

I guess the alternative - and this might be better because it covers all cases? - is that `Task` is improved to be more like `MonadError` and include things like `raiseError` and `handleErrorWith` so that when you did you `Task.Delayed` you'd put the recover code right there and that would insert the default, which would be provided in the `WebSocket.connect` method with an `onError: Msg` or `onError: Err => Msg` sort of a field.

What do you think?

gvolpe:
@davesmith00000 I would say I'd rather keep things simple. If we start adding things similar to what `MonadError` provides then I feel we would be reinventing the wheel, and we'd be better off using Cats Effect.

That's why I think either of these two proposals should suffice in this iteration.

```scala
/** A task that suspends the side-effect of producing either A on success or E on failure */
final case class Delay[E, A](thunk: () => A, onError: Throwable => E) extends Task[E, A]

/** A task that suspends the side-effect of producing A. If it fails, an error message is returned instead */
final case class Delay[A](thunk: () => A) extends Task[String, A]
```

The message handling part happens when we run the task via `Cmd.RunTask`, as in the following example.

```scala
case Msg.WebSocketStatus(Status.Connecting) =>
  (model, Cmd.RunTask(
    WebSocket.connect(
      address = model.socketUrl,
      onOpenMessage = "Connect me!",
      keepAliveSettings = KeepAliveSettings.default
    ),
    {
      case Left(err) => Status.ConnectionError(err).asMsg
      case Right(ws) => Status.Connected(ws).asMsg
    }
  ))
```

If you feel determined with the idea, though, feel free to take over because I don't know if I'm grasping what you have in mind 😄

davesmith00000:
Fair enough, I suggest that you add the `notifer` change then:

`notifier(try { Right(thunk()) } catch { case NonFatal(e) => Left(e.getMessage) })`

Leave the rest as it is, and then we can wrap up this PR. I'll review the Task stuff after that and see how strongly I feel about it.

Does that sound reasonable?

gvolpe:
Sounds good @davesmith00000 , yes, done :)

davesmith00000:
Brilliant, and thank you for all the work! ☺️

gvolpe:
@davesmith00000 before you merge, let me correct one small mistake, and then feel free to squash the commits.

tyrian/js/src/main/scala/tyrian/websocket/WebSocket.scala
```diff
@@ -4,85 +4,110 @@ import org.scalajs.dom
...
+    private var timerId = 0;
 
-  def apply(address: String): WebSocket =
-    new WebSocket(address, None)
+    @SuppressWarnings(Array("scalafix:DisableSyntax.null"))
+    def run(): Unit =
+      if socket != null && WebSocketReadyState.fromInt(socket.readyState).isOpen then
+        println("[info] Sending heartbeat 💓")
+        socket.send(msg)
+      timerId = dom.window.setTimeout(Functions.fun0(() => run()), timeout.toMillis.toDouble)
```
davesmith00000, line 110: No need to change it now, but I wonder if we can do this with a `Sub`? I'll have a look after it's merged.

PR #57, opened 2022-02-23
Fixed #55 - BUG: value attribute is unusable

### LINKED ISSUE #55: BUG: `value` attribute is unusable
If you try to set `value := "hello"` you can an ambiguous overload error.

The workaround is to use: `attribute("value", "hello")`

davesmith00000:
As reported by @gvolpe and @JonasAckermann.


hobnob (approved):
Looks good

PR #58, opened 2022-02-23
Fixed #54: 'No value' attributes with predicate

### LINKED ISSUE #54: Allow 'no value' attributes to accept a predicate
In Elm, a field like `hidden` can be expressed as `hidden(true|false)` so that you can decide whether to use the hidden attribute in-line.

My also require the production of the notion of an empty attribute.

davesmith00000:
(cc @gvolpe)

This PR allows you to:

1. Set no-value attributes based on a predicate, i.e. `hidden(true)` (hides the element, false removes the attribute showing the element)
2. Use `EmptyAttribute` or `NamedAttribute` if you wish, on top of the normal `Attribute`, `Property` and `Event` `Attr` types.

This is insanely hard to test, I did try. I have added tests for the SSR version but I ended up manually testing the dom version. Need to work out some sort of dom based testing with snabbdom. There is a ticket open for this.

### DIFF
--- a/tyrian/shared/src/main/scala/tyrian/Attr.scala
+++ b/tyrian/shared/src/main/scala/tyrian/Attr.scala
@@ -4,6 +4,16 @@ package tyrian
 sealed trait Attr[+M]:
   def map[N](f: M => N): Attr[N]
 
+/** An attribute of an HTML tag that is does not exist, used as a "do not render" placeholder
+  */
+case object EmptyAttribute extends Attr[Nothing]:
+  def map[N](f: Nothing => N): EmptyAttribute.type = this
+
+/** An attribute of an HTML tag that only has a name, no value
+  */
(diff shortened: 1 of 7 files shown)


tyrian/shared/src/main/scala/tyrian/Attr.scala
```diff
@@ -4,6 +4,16 @@ package tyrian
 sealed trait Attr[+M]:
   def map[N](f: M => N): Attr[N]
 
+/** An attribute of an HTML tag that is does not exist, used as a "do not render" placeholder
```
JonasAckermann, line 7:
```suggestion
/** An attribute of an HTML tag that does not exist, used as a "do not render" placeholder
```
? not sure if that was the intended meaning

gvolpe:
This is awesome! Thanks a lot for the quick turnaround 😻

PR #59, opened 2022-02-23
Fixed #56: Create a "focus" Cmd

### LINKED ISSUE #56: Create a "focus" `Cmd`
Consider replicating some of the behaviours in this Elm package, particularly the `focus` command:

https://package.elm-lang.org/packages/elm/browser/latest/Browser.Dom

davesmith00000:
(cc @gvolpe)

This PR gives you commands to focus and blur (un-focus?) HTMLInputElements.

They are based on the commands in the Elm [Browser.Dom](https://package.elm-lang.org/packages/elm/browser/latest/Browser.Dom) package.

I have an open question about this implementation: Currently it insists that you deal with the possible error (i.e. element not found) and even the success of running the side effect by producing more messages. This might be desirable, but I wonder if it would be good to somehow have a fire and forget version, and maybe a version that only emits an event on error (if it's even possible to build such things)? I'll sleep on it.

### DIFF
--- a/sandbox/src/main/scala/example/Sandbox.scala
+++ b/sandbox/src/main/scala/example/Sandbox.scala
@@ -2,6 +2,7 @@ package example
 
 import tyrian.Html.*
 import tyrian.*
+import tyrian.cmds.Dom
 import tyrian.cmds.Logger
 import tyrian.websocket.*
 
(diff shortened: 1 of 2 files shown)


hobnob:
I think an error may be preferable - I think it might be confusing to a developer if the `focus` command wasn't working, but the code looked fine, only to find a silly spelling mistake

hobnob (approved):

gvolpe:
Awesome @davesmith00000 ! I agree with @hobnob here, I'd rather have error raised than swallowed.

PR #69, opened 2022-02-26
Fixed #60: Values are properties, not attributes

### LINKED ISSUE #60: HTML input value is not reflected on the view
It seems the `value` attribute of an input field is not correctly reflected on the view when the model changes.

```scala
case class Model(input: String)

input(`type` := "text", attribute("value", model.input))
```

The `input` field is updated either on `Msg.SymbolChanged` or reset (`input = ""`) on `Msg.Subscribe`.

```scala
// view
input(`type` := "text", attribute("value", model.input), onInput(s => Msg.SymbolChanged(s)))

// update
case Msg.SymbolChanged(in) =>
  model.copy(input = in) -> Cmd.Empty

case Msg.Subscribe =>
  val nm  = model.copy(input = "")
  val cmd = model.ws.map(ws => Cmd.Batch(ws.publish(WsIn.Subscribe.asJson.noSpaces), refocusInput))
  nm -> cmd.getOrElse(Cmd.Empty)
```

Everytime the model is updated in the first message, I can see the HTML input value is updated (seeing this by inspecting the element in the browser). However, when the `Subscribe` message is handled and the `input` in the model is reset, the `value` attribute is properly reset as well, but the view remains with the previous value.

Here's an example from the trading application, right after clicking subscribe, where I first noticed the issue.

issue

The `value` attribute is reset but the view still shows the previously entered value "CHFEUR". I'd expect the input field to be cleared up as well.

davesmith00000:


davesmith00000:
(cc @gvolpe and @aumann)

aumann (approved):
lgtm

PR #70, opened 2022-02-26
Some progress towards fixing mdocs

armanbilge:
So, this helps things but it's still broken.

If you remove the `cmd.md` page then the docs build fine. In fact, including _only_ the following block is sufficient to break it.

```
```scala mdoc:js:shared
import tyrian.*
import tyrian.cmds.*
``` 

The stack trace points us to here, which was recently modified. Possible mdoc bug?
https://github.com/scalameta/mdoc/blob/9014e2d72db7e6c8def870bfbd4f53999582b024/mdoc-js/src/main/scala/mdoc/modifiers/JsModifier.scala#L163

```
[info] mdoc.internal.markdown.ModifierException: mdoc:js exception
[info] Caused by: java.util.NoSuchElementException
[info]  at java.base/java.util.HashMap$HashIterator.nextNode(HashMap.java:1513)
[info]  at java.base/java.util.HashMap$ValueIterator.next(HashMap.java:1539)
[info]  at mdoc.modifiers.JsModifier.postProcess(JsModifier.scala:163)
[info]  at mdoc.modifiers.JsModifier.postProcess(JsModifier.scala:131)
[info]  at mdoc.internal.markdown.Processor.processDocument$$anonfun$3$$anonfun$1(Processor.scala:62)
[info]  at mdoc.internal.markdown.Processor.runModifier(Processor.scala:72)
[info]  at mdoc.internal.markdown.Processor.processDocument$$anonfun$3(Processor.scala:63)
[info]  at scala.runtime.function.JProcedure1.apply(JProcedure1.java:15)
[info]  at scala.runtime.function.JProcedure1.apply(JProcedure1.java:10)
[info]  at scala.collection.immutable.List.foreach(List.scala:333)
[info]  at mdoc.internal.markdown.Processor.processDocument(Processor.scala:64)
[info]  at mdoc.internal.markdown.Markdown$.toMarkdown(Markdown.scala:137)
[info]  at mdoc.internal.cli.MainOps.handleMarkdown(MainOps.scala:83)
[info]  at mdoc.internal.cli.MainOps.handleFile(MainOps.scala:111)
[info]  at mdoc.internal.cli.MainOps.$anonfun$1(MainOps.scala:157)
[info]  at scala.collection.LinearSeqOps.foldLeft(LinearSeq.scala:169)
[info]  at scala.collection.LinearSeqOps.foldLeft$(LinearSeq.scala:165)
[info]  at scala.collection.immutable.List.foldLeft(List.scala:79)
[info]  at mdoc.internal.cli.MainOps.generateCompleteSite(MainOps.scala:159)
[info]  at mdoc.internal.cli.MainOps.run(MainOps.scala:178)
[info]  at mdoc.internal.cli.MainOps$.process(MainOps.scala:270)
[info]  at mdoc.Main$.process(Main.scala:26)
[info]  at mdoc.Main$.process(Main.scala:21)
[info]  at mdoc.Main$.main(Main.scala:16)
[info]  at mdoc.Main.main(Main.scala)
[info] info: Compiled in 11.84s (1 error)
```


armanbilge:
Hmm. I was able to fix the docs completely by rolling back to mdoc 2.3.0 (as well as SJS to 1.7.0 as required by mdoc 2.3.0). So it seems like this is a regression actually, sorry @keynmol.

build.sbt
```diff
@@ -148,15 +148,16 @@ lazy val jsdocs =
     .settings(
       neverPublish,
       organization                           := "io.indigoengine",
-      libraryDependencies += "org.scala-js" %%% "scalajs-dom" % Dependancies.scalajsDomVersion
+      libraryDependencies += "org.scala-js" %%% "scalajs-dom" % Dependancies.scalajsDomVersion,
+      scalaJSLinkerConfig ~= { _.withModuleKind(ModuleKind.CommonJSModule) }
     )
+    .dependsOn(tyrian.js)
+    // .dependsOn(tyrianIndigoBridge.js)
```
davesmith00000, line 155: I'd really like to switch this for `"io.indigoengine" %%% "tyrian" % tyrianDocsVersion` ideally, the current in dev version is the one I want the docs checked against. Feels like that should work but it threw up new problems for me.

davesmith00000: Ah but I guess that won't work because they're all published against a later version of Scala.js.

davesmith00000 (commented):
Thanks for investigating. 😄 

The annotation changes you've made are all ones I've made before - so nice its to know I wasn't barking up the wrong tree completely. 😅 

I guess I'll keep this open and leave the current docs as they are while we see if the regressions really is a regression and if it can be fixed.

Thanks again!

build.sbt
armanbilge: Yeah, I wouldn't take any of the changes in this PR too seriously :) mainly, it's to demonstrate that things _should_ work and hopefully help track down that mdoc regression in the latest release.

docs/concepts/cmd.md
```diff
@@ -38,7 +38,7 @@ Commands can be produced as part of a result of calling the `init` or `updateMod
 
 Here is an example in which, on receiving a message `Msg.LogThis`, we are not going to change the model, but we want to write to the browser's JavaScript console:
 
-```scala mdoc:silent
+```scala mdoc:js:shared
```
keynmol, line 41: Copying from Discord for posterity:

That file only has `:shared` and `:compile-only` modifiers - which means that there's no entry points to actually run the JavaScript code - which means it's probably being eliminated by the linker - which means it produced no JavaScript files, which in turn crashes the modifier.

Just changing one of the blocks from `mdoc:js:compile-only` to `mdoc:js` actually fixes the generation with Mdoc 2.3.1 and Scala.js 1.9.0.
I changed this one: https://github.com/PurpleKingdomGames/tyrian/pull/70/files#diff-c35eaeeeac3890e0c4181f6aebc4d5048b37049123f63dd434b48191d75dfad9R58

davesmith00000:
Thanks chaps! I believe I have now fixed this on the main branch. The comments you left were exactly what I needed. 🙏

PR #72, opened 2022-02-27
docs: remove main module initializer option

gvolpe:
It seems no longer necessary and the examples don't compile with this option, so removing it from the docs.


davesmith00000 (approved):
Thanks!

PR #77, opened 2022-02-28
Fixed #45: Support hot-reloading

### LINKED ISSUE #45: Dev Mode: Store model state on update
If we allowed a dev mode that stored the model state on update, than a hot reload would not reset the page to the beginning.

davesmith00000:
This totally works ...but is a fairly disappointing solution that once again highlights that I need to rethink `Task`s / `Cmd`s / `Sub`s at some point.

I was hoping to to produce a subscription that just did the snapshotting for you on top of what we have here, but it proved to be a real pain.

None the less, with these two commands and a little subscription, you can always carry on from where you left off now.

### DIFF
--- a/sandbox/src/main/scala/example/Sandbox.scala
+++ b/sandbox/src/main/scala/example/Sandbox.scala
@@ -7,14 +7,21 @@ import tyrian.cmds.LocalStorage
 import tyrian.cmds.Logger
 import tyrian.websocket.*
 
+import scala.concurrent.duration.*
 import scala.scalajs.js.annotation.*
 
 @JSExportTopLevel("TyrianApp")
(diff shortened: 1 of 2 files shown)


sandbox/src/main/scala/example/Sandbox.scala
```diff
@@ -157,6 +182,8 @@ enum Msg:
   case Log(msg: String)
   case WebSocketStatus(status: Status)
   case Clear
+  case OverwriteModel(model: Model)
+  case TakeSnapShot
```
aumann, line 186: nitpicking: maybe TakeSnapshot (lower case shot) for consistency (compare HotReload.snapshot)

davesmith00000: Done, thanks.

PR #81, opened 2022-03-01
WebSocket: keep-alive mechanism as a Sub

### LINKED ISSUE #67: WebSockets: Can the heartbeat be a form of `Sub`?
The heartbeat is currently using a `setTimeout` right in the WebSockets code.

gvolpe:
closes #67 

Not sure if this is what you had in mind, but I went for it anyway 😄 In addition, I added a graceful shutdown example. 

👍🏽 re-using the `Sub.every` mechanism.
🤔 the heartbeat message needs to be sent by the user.

I think it would be great if we could re-use `Sub.every` but the heartbeat message was sent internally, as this is client-server communication that has nothing to do with user messages.

I'd love to hear what you think! 😃

### DIFF
--- a/sandbox/src/main/scala/example/Sandbox.scala
+++ b/sandbox/src/main/scala/example/Sandbox.scala
@@ -133,6 +133,10 @@ object Sandbox extends TyrianApp[Msg, Model]:
           )
         )
 
+      case Msg.WebSocketStatus(Status.Disconnecting) =>
+        println("Graceful shutdown of WS connection")
+        (model.copy(echoSocket = None), model.echoSocket.map(_.disconnect).getOrElse(Cmd.Empty))
+
       case Msg.WebSocketStatus(Status.Disconnected) =>
         println("WebSocket not connected yet")
         (model, Cmd.Empty)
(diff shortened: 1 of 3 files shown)


davesmith00000:
I think it looks great! ...and it's pretty much what I had in mind. 😄 

My feeling is that we could safely merge it. It's possible that once people try using it they may have feedback about whether or not they like the pattern... but that's ok! I'm sort of finding that sooner or later, everything is a `Sub` or a `Cmd`, and as a user you tend to see everything... and I kinda don't mind. 🤷 

Are you happy with it? The way it works / how it feels? If so I'll hit the merge button. 👍 

Thanks!

gvolpe:
Great! I'm willing to give it a try. I felt it a bit awkward at first, but after a while I think I could get used to it. The heartbeat / keep-alive mechanism is not a standard in WebSockets, so it's okay it needs to live in user land.

This implementation also simplifies the `KeepAliveSettings` as we no longer need the `enabled: Boolean` and `message: String`. Instead, we can handle these two cases directly in the updates. 

```scala
// to disable
case WebSocketEvent.Heartbeat =>
  model -> Cmd.Empty
  
// to enable
case WebSocketEvent.Heartbeat =>
  model -> model.ws.map(_.publish("<heartbeat>").getOrElse(Cmd.Empty)
```

Go ahead with the merge! 💪🏽

davesmith00000 (approved):
Tested, works great. I'll try and write some docs for this while I'm upgrading the example project ahead of the next release. Thanks again! 😄

PR #84, opened 2022-03-04
Improve `WebSocket`s example ahead of release

davesmith00000:
This work was started in preparation for the next release of Tyrian.

The `WebSocket` implementation has had a lot of work thanks to @gvolpe, and I knew there was going to be a lot of changes to make and land before the release.

The sandbox (a test project) had been updated to include the new changes, but it is quite messy in there (by its nature as as a smorgasbord of random functionality). So I first moved everything over to the existing websocket example, then I started cleaning it all up.

There are three things to look at here:

1. I've changed the `WebSocket.connect` methods to return a `Cmd[Msg]`, as I think the result is cleaner. The `Task` versions still exist in parallel as `connectTask`.
2. The example code base is now quite different from the `Sandbox` version. I think it's cleaner, but if anyone can see any other ways to further tidy it up and/or improve the `WebSocket` APIs themselves - speak now.
3. Part of the clean up resulted in making an `EchoSocket` ...thing. I'm wondering if this is a pattern work generalising?


hobnob (approved):

gvolpe:
That's very nice! I would make `connectTask` private, otherwise that would be confusing to users. Not sure if there's any value in making it public?

davesmith00000:
Agreed. I hadn't quite decided if they had any value being public, but I think you're right so I've made them private now, thanks.

PR #85, opened 2022-03-07
nix: add correct mill overlay

gvolpe:
Tiny change that overrides the `jre` version used by `mill` in the right way.


davesmith00000 (approved):
Thanks!

PR #86, opened 2022-03-28
Replace Task with Cats Effect 3

davesmith00000:
# What is this?

The `Task` implementation that Tyrian inherited from Scalm - despite being minimal - was pretty much fit for purpose. The JS runtime is not the JVM*, and the task implementation had the benefit of being pretty darn simple.

(* I am no expert on effect monad building or of comparing runtime environments, but even I can see that there are different concerns at play here, unless I'm very much mistaken...)

Swapping `Task` for `IO` has not been a fun exercise, but I've persisted. I don't actually think Tyrian is (yet) any better off for this change (and is arguably a good deal more complicated now) except hopefully eventually it will be in three specific ways.

1. Principle of least surprise - `Task` was an effect monad - no getting away from it - but it also didn't always behave in the expected way, and that was a bit confusing/frustrating at times.
2. Familiarity - Hopefully at the end of this process, you will be able to take your working CE3-compatible-effect-monad-of-choice knowledge and apply it here. It ought to work more or less like it would anywhere else.
3. Ecosystem - I suspect that _right now_, there is no significant CE3 (compatible) library ecosystem for Scala.js. But if there's going to be a Scala.js ecosystem at all for the pure FP enthusiasts, this is probably where it'll be (I think).

## Steps...?

- [x] Step 1: Swap `Task` for `IO` (just make it work...)
- [X] Step 2: Replace `IO` with the cats effect kernel typeclasses
- [X] Step 3: Reconsider how the runtime works
- [ ] ~~Step 4: Is it possible to reduce the amount of times I have to provide the full type params in sandbox?~~
- [X] Step 5: Make TyrianApp it's own project called `tyrian-io`
- [ ] ~~Step 6: Consider `tyrian-zio` and `tyrian-monix` if they have JS runtimes?~~ These libraries don't yet support Scala.js + Scala 3 + CE3

## Known things to fix/review...

- [x] Make the shape of the result types consistent across `Cmd`s / `Sub`s
- [x] Scaladocs... everywhere...
- [x] See if it works with all the examples
- [x] Might be good to offer standard encodings for things like callbacks, and maybe nicer construction options of `Sub`s.
- [x] Revisit the name 'observable' everywhere...
- [x] Ensure the final API's remain friendly and quick to get up to speed with.
- [x] Check for TODO's!

## Testing to add - may need JSDom...

- [ ] `RandomTests` - bring back to life
- [ ] `CmdHelper` - `cmdToTaskList` (rename to flatten?)
- [ ] `SubHelper` - `flatten`, `aliveAndDead`, `findNewSubs`, and `toRun` too maybe?
- [ ] `Sub` removal when dead

### Step 1

I'm not saying this is good code, and clearly I've just commented out loads of tests, and I haven't tested everything, and it's horribly inconsistent, and it's probably not good CE3 code... in fact it's kind of a mess... but it works. In this first step, I've tried not to deviate from the original runtime design too much for now to reduce moving parts.

### Steps 2 & 3

So it's turns out, these were in the wrong order. I had a big false start and spent a week banging my head off my monitor. Then in one night I threw all my work away, re-worked the runtime moving the effects to the edge, and swapping in `Async` for `IO` was then a piece of cake. The runtime is still very side-effect-y... but I don't care.

### Step 5

This was nice and easy, I moved the `TyrianApp` to a `tyrian-io` artefact that I'll publish separately. This means `tyrian` itself (and the bridge) are only dependant on the cats effect kernel - theoretically able to support any compatible effect type. I did a small enhancement makeing a `TyrianAppF` trait that the `IO` bound version extends. This means for any future effect monad implementations, we only have to add a single new effect-type-bound module that is just one function long.

In practice (at the time of writing), the only effect monad that is published for CE3 + Scala 3 + Scala.js _is_ `IO`. But I'm sure that'll change eventually.

### (Step 6 - not needed)


davesmith00000:
(FYI @gvolpe)

davesmith00000:
I simultaneously would like this over and done with, but also it's a bit of a learning journey for me so I don't want too much assistance at the moment. I'll reach out to someone if I get stuck, or if I think I'm done and want a second pair of eyes from someone who knows what they're doing.

davesmith00000:
Thanks @armanbilge. 👍 

Bit of crossed wires there due to me not being clear originally, sorry about that. I'm personally not interested in an ecosystem of things. However, one of the arguments people have made to me, in support of the case for Cats Effect in Scala.js, is that there is some ecosystem of libraries they can take advantage of. I have no idea what that would be in the browser, and it leads me to suspect that they really mean for use with Scala.js based AWS Lambdas or something. But who knows. 🙂

davesmith00000:
I've broken the back of this now. Finally. It's been a very unpleasant week.

davesmith00000:
# Quick migration notes:

## Deps

Update:
`"io.indigoengine" %%% "tyrian" % tyrianVersion` -> `"io.indigoengine" %%% "tyrian-io" % tyrianVersion`

Add:
`"org.typelevel" %%% "cats-effect" % "3.3.8"`

## New imports needed

`import cats.effect.IO`

## Syntax/API changes:

`Cmd.Empty` -> `Cmd.empty`
`Sub.Empty` -> `Sub.empty`

`Cmd.SideEffect { () => println("hello") }` -> `Cmd.SideEffect { println("hello") }`

`resultToMessage` used to be: `Either[Error, Value] => Msg`
Most are now: `Value => Msg`
The expectation is that most 'values' are now enums / ADTs / your type of choice.

`Task` removed completely. Use `Cmd.Run` to run some `IO`.

`Sub.ofTotalObservable` and `Sub.OfObservable` removed. Use `Sub.Observe`. This takes a `String` id, a `toMsg: A => Msg` and an 'observable'. The observable is really a callback with a function that has a way to cancel/close/remove the thing producing events.

The type is: `F[(Either[Throwable, A] => Unit) => F[Unit]]` where the `(Either[Throwable, A] => Unit)` is a callback, and the `F[Unit]` is the cancel function.

Example from the Mario demo making a `Sub[IO, Double]` for listening to `requestFrameAnimation`:

```scala
val observable =
  IO.delay { (callback: Either[Throwable, Double] => Unit) =>
    var handle = 0
    def loop: Double => Unit = time => {
      callback(Right(time))
      handle = dom.window.requestAnimationFrame(loop)
    }
    handle = dom.window.requestAnimationFrame(loop)
    IO.delay(dom.window.cancelAnimationFrame(handle))
  }

Sub.Observe(
  "requestAnimation",
  observable,
  identity
)
```

## Type annotations:

You'll see this a lot... you need to add type annotations.
`Error: "Could not find an instance of Monad for F"`

Examples:

- `Cmd[Msg]` -> `Cmd[IO, Msg]`
- `Sub[Msg]` -> `Sub[IO, Msg]`
- `Sub.every(1.second, "clock-ticks").map(Msg.apply)` -> `Sub.every[IO](1.second, "clock-- ticks").map(Msg.apply)`
- `TyrianIndigoBridge[A]` -> `TyrianIndigoBridge[IO, A]`
- `TyrianSubSystem[A]` -> `TyrianSubSystem[IO, A]`
- `Sub.fromEvent[KeyboardEvent, KeyboardEvent]` -> `Sub.fromEvent[IO, KeyboardEvent, KeyboardEvent]`
- `WebSocket` -> `WebSocket[IO]`
- `Logger.info` -> `Logger.info[IO]`

davesmith00000:
Hi @armanbilge and @gvolpe,

No pressure, there is a lot here, but if you felt like looking over this for any obvious improvements I could make, now would be a good time as I won't get to work on this for a couple of days (probably).

I've got some more testing to add / fix, but I think its otherwise nearly done. A fresh pair of eye's couldn't hurt! I've added some migration notes as a comment so you can see what the changes mean to the end user currently.

Thanks in advance - again if you're busy don't worry. 😄

davesmith00000:
Thanks @armanbilge - there's some stuff to ponder in there and I'll have to come back later when I have some more time.

I did try a range of formulations of this - in an amateurish, cargo-cult-y, bumbling sort of way. 😄  It's not lost on me that my encoding of the callback is very similar to `Async[F].async` for example. There are problems though, off the top of my head:

1. The _current_ lifecycle of the runtime doesn't map neatly onto how you might like to use something like fs2 (I think), and re-organising it is probably a job for a separate PR.
2. Many of the examples I've seen use things like `Resource`, but that doesn't have the right lifecycle either, it opens, uses, and closes (or uses forever). What I need is open, use for a while, maybe forever, maybe close later when _I_ want to not (only) when some notional fiber is cancelled.
3. I realise I could use less powerful typeclasses, but got bored of the compiler blowing up while I was working it out, should probably revisit now it's stable. 👍  (I like the idea of being able to bring back pure elements, not sure how that plays with the Cmd/Sub flattening process. Might be fine? We'll see.)

The problem with the current runtime (which works perfectly well, but is just a bit OO-y) eventually is that a msg callback basically re-triggers the whole process (unlike Indigo, which is simpler and stricter) so you end up with a nested situation where you need to run the thing from inside yourself. Ah ha! I hear you cry! Just flatMap and pass the thing to run back out! Sounds right, then you need something like a mutable channel or queue you can keep feeding messages into, and then you need to run again whenever there is more to do. I feel certain this must be doable (without polling? surely?) but again, not a use case I could find a reasonable example of. There are _endless_ examples of fs2/cats-effect running a queue where people offer and take to/from it _within the same for-comp_ (when is that noddy example really useful I wonder?), but nothing I could see about doing that out of band. I'm sure you can do that, but I need a lot more free time to work out how to do it. It feels like another documentation gap. I can see from all the docs how to use fs2 linearly - piece of cake - but I'm struggling to find good references to how one uses it as the basis of a long running stateful application. Again I know it must be possible because Http4s does it right? But the Http4s code base is not a friendly reference point for such things.

Most of this boils down to me being too time poor to just sit and absorb the available tools in all their depth and glory, I realise. And as I result I find I lose my usual optimistic disposition and end of moaning a lot. I manage to find a couple of hours a day to myself - max - and I want to make progress in every minute, which makes the brick-wall learning curve of CE/FS2 for lib/application building... off-putting ..I'll try and stop moaning. 😅

davesmith00000:
I'm leaving that comment there for posterity, but retracting it. I feel like I'm just moaning. I appreciate the feedback @armanbilge - thank you.

armanbilge:
Oh, I was just finishing up a response 😅 fwiw I thought it was all reasonable :) This was what I was going to say.

---

Yep, I assumed it had a lot to do with the runtime which I fortuitously avoided in my review 😉 

I guess focusing specifically on this:
```scala
val observable: F[(Either[Throwable, A] => Unit) => F[Unit]] = ???
```

I assume you use it roughly something like this:
```scala
val myCallback: Either[Throwable, Whatever] => Unit = ???
for
  subscribe <- observable // get the subscription
  cancel <- F.delay(subscribe(myCallback)) // start the subscription with the callback
  // do stuff
  _ <- cancel // ok, time to cancel it
yield ()
```

I think this is roughly equivalent to:
```scala
val observable: Stream[F, A] = ???
val myCallback: Either[Throwable, A] => Unit = ???
for
  // get a stream of errors or As, invoke the cb for each, run this in a fiber
  fiber <- observable.attempt.foreach(x => F.delay(myCallback(x))).compile.drain.start
  // do stuff
  _ <- fiber.cancel
yield ()
```

The point being, even if the runtime needs another PR(s) worth of iteration, hopefully user APIs don't need to know about that :)

> I'm struggling to find good references to how one uses it as the basis of a long running stateful application.

Yeah, that's a good question. I wonder if this would be a helpful example for you.
https://github.com/djspiewak/kitteh-redis
But also, you should ask on the Discord :)

> absorb the available tools in all their depth and glory

Yes, this really just takes time and practice and code reviews which might put you into to new and better patterns for common problems. They are overwhelmingly powerful APIs, particularly fs2 and it's just impossible to fit it all in your head 😅

gvolpe:
I will have a proper look shortly, but I feel a lot of comments from @armanbilge are spot on. I can try checking out the branch and see if I can help with the observable design, that's an interesting one :)

I learned a bit by using both Elm and Tyrian, so perhaps I can offer some help.

davesmith00000:
Regarding:
> I assume you use it roughly something like this:

So following the comments in the code example there, you've got:

1. Grab the subs
2. Run them
3. Do some stuff to get a result
4. Cancel.

And that all happens in one go.

But that's not what Tyrian actually does. It does (1) and (2), then the doing stuff happens until the user decides to stop them by removing the sub, in the meantime the 'cancel' operation has been stored so that if and when the user does remove it, it can be invoked.

Contrived example:

- A web app developer attaches a mouse move listener to the window in order to draw a custom cursor or something.
- In the app's options, the end user decides to disable custom cursors (maybe the option is the for performance reasons), so the app's code removes the mouse move sub from the subs list based on the options stored in the model state, i.e. it just returns a different / changed list of subs.
- The subs list (`def subscriptions`) is stateless, which means that in the runtime every time the model changes, we have to work out a diff of new subs in the list versus known active subs (which are stored mutably), and call the cancel function (which we also stored away as `sub id: String => cancel: F[Unit]`) on any running subs that are no longer in the active list.

Does that help?

armanbilge:
Right, sorry, my "do stuff" was a bit too vague but basically you can do whatever you want there. E.g
```scala
val observable: Stream[F, A] = ???
val myCallback: Either[Throwable, A] => Unit = ???
for
  fiber <- observable.attempt.foreach(x => F.delay(myCallback(x))).compile.drain.start
  _ <- storeSubCancel(key, fiber)
  // do stuff
yield ()
```

Somewhere else in the program:
```scala
for
  // doing stuff
  // oops, time to cancel it!
  fiber <- getSubCancel(key)
  _ <- fiber.cancel
yield ()
```

davesmith00000:
> Yeah, that's a good question. I wonder if this would be a helpful example for you.
> https://github.com/djspiewak/kitteh-redis

That is very good, yes. Thanks for sharing. 🙏 

The data flow is pretty much what I had in mind when I was scratching around, and is really very close to what I need. I don't understand all of it yet (just reading on github), but the missing puzzle piece I couldn't find in my own earlier attempts is somewhere in the [`Network` bit](https://github.com/djspiewak/kitteh-redis/blob/main/src/main/scala/kitteh/Server.scala#L281). Essentially where the `Network` keeps listening for events in order to feed them (via the command eval) back into the main stream to the acted upon _after_ the main resource / stream has been initially evaluated to produce the server.

My suspicion is that I can make something very similar once I understand it, it's just that I'll need to break the current runtime organisation to do so by essentially having all listeners post to a sort of proxy which then feeds into the stream.

I'll clone the repo later and have a proper poke around with my IDE. 👍

armanbilge:
> having all listeners post to a sort of proxy which then feeds into the stream.

Sounds like the infamously undocumented Channel 😆

davesmith00000:
Update:
Tyrian is now using FS2 under the hood, I've taken on a bunch of the review points, particularly the one about making the `F` in `Cmd`/`Sub` covariant - much better.
`Cmd`s and `Sub`s are now evaluated in the stream having been posted to a `Channel`, however, `Sub` is still it's new 'observable' shape - so that's the next (dare I say it... final?) biggish thing to work on.

tyrian/js/src/main/scala/tyrian/runtime/TyrianRuntime.scala
```diff
@@ -47,63 +47,92 @@ final class TyrianRuntime[F[_]: Concurrent, Model, Msg](
...
+        n <- vnode.get
+        _ <- vnode.set(n match {
+          case Some(existingNode) => Some(render(existingNode, updatedState))
+          case None               => Some(render(node, updatedState))
+        })
+        sideEffects <- gatherSideEffects(cmd, subscriptions(updatedState))
+      } yield Stream.emits(sideEffects)
+
+    Async[F].flatMap(results) { (stream: Stream[F, F[Unit]]) =>
+      stream.foreach(channel.send(_).map(_ => ())).compile.drain
```
gvolpe, line 84: `.map(_ => ())` can be replaced by `.void` (comes from the `Functor` syntax).

gvolpe (commented):
Looking good so far! 

@davesmith00000 do you think this could already be reviewed? I am so sorry I couldn't yet make time to check this out  (and I'm going on holidays next week 🙈 ) but I'd definitely like to help out here! 😃

tyrian/js/src/main/scala/tyrian/runtime/TyrianRuntime.scala
davesmith00000: Neat! I've changed that now.

davesmith00000:
No worries @gvolpe, don't sweat it. 😄 

Although it's been a struggle, I'm grateful for the learning opportunity. I'll see how I get on with the subs bit and give you a nudge if I start clawing my eyes out. 😉 

In terms of can it be reviewed, I think ...nearly ...cautiously yes? There a few things to look at:

1. I haven't finished testing it with all the usual projects so, I need to do that.
2. I'd like to try and write some tests for the runtime. I've been flying by the seat of my pants for too long...
3. I'd like to casually walk around and do some clean up, such as seeing if I can reduce the typeclasses a bit further, and there's still a couple of `var`s in the runteim that could _maybe_ be `Ref` instances instead (if it's not too much of a PITA).
4. Subs: I'm going to re-review Arman's comments and spend some time thinking about a better solution. (I may not find one)
5. Anecdotally, it feels slower. Not the general page rendering or interactions stuff but the command / subscription execution. I haven't measured it, and I'm confident that the new version is more correct, but it feels a fraction more sluggish. I'm wondering if it's something to do with how often items are pulled from the `Channel` instance or something? Need to look into it.

davesmith00000:
Regard the performance drop (point 5): The issue goes away with a `fullOpt`, so it'll be slightly slower during dev but in production it should be nice and snappy.

armanbilge:
Just wanna pause and say how incredible watching this transformation has been :') really amazing work. Building Tyrian on CE/fs2 is truly a feather in their cap.

davesmith00000:
Subscriptions. Hmmm. Yes. Right.

I think I see what you're saying now @armanbilge.

In the madness of just trying to port subscriptions from `Task` to `IO`, I was a bit preoccupied with how they work _now_ and not thinking about what that _are_.

`Sub`s are signals, a value that changes over time, but in practice (to borrow FRP vernacular) they are _discrete_ not _continuous_ because ultimately they are based on events fired by the browser (sooner or later), as a pose to a value you can sample at any moment.

Therefore, in practice, subscriptions are streams. If subscriptions are streams and the whole thing is running on fs2, then we might as well model them that way.

I feel like I probably know how I'd just blend in a bunch of fixed streams, the bit I'm not sure about is dynamically adding and removing streams... might be easier than I think.

Example:

1. You have a sub in the subscriptions list that throws out a `ping` message every second. The list is re-evaluated every time a msg passes through the runtime. **(So we must create the stream, and not re-create it when the list is re-evaluated)**
2. At some point, the app's model changes, and the web app dev ~removes~ decides not to include that sub in the subscriptions list any more. **(So we must close the stream down)**
3. The model changes a second time, and the dev includes the sub in the list again. **(So we must re-create/restart the stream)**

I'll have a play.

davesmith00000:
(cc @gvolpe - thoughts and comments welcome!)

armanbilge:
> the bit I'm not sure about is dynamically adding and removing streams...

I _think_ I just did something like this myself.

Rather than think about it as a list of `Stream`s, it may be helpful to think about it as a pool of `Resource`s. i.e., you can run a `Stream` as a fiber (`.complie.drain.background`) which you dispose of by cancelling.

In my case, I had:
- a stream representing a dynamic list of keys `Stream[F, List[K]]` 
- a function `K => Resource[F, HTMLElement]` (where in practice, that resource is also managing a `Stream` running in the background)

And I wanted to transform it to a `Stream[F, List[HTMLElement]]`. Specifically, I wanted it to re-use existing `Element`s where possible, and only create/dispose elements when the keys changed.

In any case, the idea is to maintain a pool of key-value pairs, where the value also includes the finalizer to dispose of that key. So as the dynamic list of keys updates, you can check if it's already in the pool. If not, allocate. If there's anything in the pool you _don't_ want anymore, dispose.

Not sure if it's any help, but my ~first~ second sketch looks like this.
https://github.com/armanbilge/calico/blob/main/calico/src/main/scala/calico/dsl.scala#L292-L316

davesmith00000:
That is actually very close to how it works right now if I'm following you correctly (ignoring that it's IO's not resources of streams), which sounds hopeful. The pool of resources is a good thought. 🤔

gvolpe:
While I am not that familiar with the concrete implementation of `Sub`, judging by your description:

> `Sub`s are signals, a value that changes over time, but in practice (to borrow FRP vernacular) they are discrete not continuous because ultimately they are based on events fired by the browser (sooner or later), as a pose to a value you can sample at any moment.

This could be modeled using Fs2's signals; more specifically, a `SignallingRef`.

So basically, adding, removing, and updating `Sub`s becomes as simple as reflecting these changes on a `Ref`, as you'd do with simple data. Clients would get a `discrete` stream of changes from these subscriptions, powered by the same data structure.

I use something like this to handle the total number of Web Socket connections, so every client knows how many clients are connected at the moment they connect.

https://github.com/gvolpe/trading/blob/main/modules/ws-server/src/main/scala/trading/ws/WsConnections.scala#L19

davesmith00000:
After further consideration, I've decided not to pursue the idea of `Sub` being an `fs2.Stream`, for principally two reasons:

1. It's overkill. Tyrian apps are _not_ based on FRP, and merging complex signals to produce the outcome isn't how they are supposed to work. Mostly all you're ever doing is listening to browser events. If a use-case emerges where it would be helpful to be able to, say, merge values/streams, then we can revisit.
2. That being the case, I'd rather not expose streams on the API, as it's an un-necessary complexity / learning hurdle.

What I have done is re-assess how Sub's are formulated somewhat. The 'observable' part is now very similar to a cats effect `Async.async`:

```scala
F[(Either[Throwable, A] => Unit) => F[Option[F[Unit]]]]
```

This is an effect that encapsulates a callback and an optional cancellation. Pretty similar to the old version, but it allows some interesting new construction possibilities that can feel somewhat like a cats effect `Resource`:

```scala
def make[F[_]: Sync, A, Msg, R](id: String)(acquire: (Either[Throwable, A] => Unit) => F[R])
  (release: R => F[Unit])(toMsg: A => Option[Msg]): Sub[F, Msg]  = ???
```

There are various permutations of that in the latest commit, such as this one used to construct `Sub.every`:

```scala
/** A subscription that repeatedly produces a `msg` based on an `interval`. */
def every[F[_]: Sync](interval: FiniteDuration, id: String): Sub[F, js.Date] =
  Sub.make[F, js.Date, Int](id) { callback =>
    Sync[F].delay {
      dom.window.setInterval(
        Functions.fun0(() => callback(Right(new js.Date()))),
        interval.toMillis.toDouble
      )
    }
  } { handle =>
    Sync[F].delay(dom.window.clearTimeout(handle))
  }
```

I did consider using a `Resource` but decided against it. I don't know if this is _right_ but the idea was to expose the ability for people to use their favourite effectful monad, not to litter the API with CE types (unless the value was clearly there).

armanbilge:
👍 on the `Stream` stuff, I especially agree about Tyrian not being FRP and that its presence in the API being unnecessarily opinionated/complex/confusing. Less sure about skipping on `Resource`, but anyway :)

FWIW personally I'm still having a really hard time wrapping my head around this:
```scala
F[(Either[Throwable, A] => Unit) => F[Option[F[Unit]]]]
```

I feel like it's isomorphic to ... `F[A]` :)

As an example, `every` could be expressed as:

```scala
Temporal[F].sleep(interval) >> Temporal[F].realTimeDate
```

Then, the `Sub` implementation can `.start` it. will get a fiber that it can `.join` to wait for the result or `.cancel` at will. In this case, the `join` is equivalent to your callback I believe, and the `cancel` is the cancellation.

> The 'observable' part is now very similar to a cats effect `Async.async`.

I do see the connection. I guess I'm confused why a user shouldn't just use `Async.async { ... }` to create a (cancelable) `F[A]`, and directly pass that to the sub, which can run/cancel at will? And `async` would only be necessary for the most complex case of course. In some cases (such as `every`) it may be possible to accomplish what's needed using simpler combinators (such as `sleep` and `realTimeDate` without ever needing to interact with a raw callback or require a strong constraint like `Sync` or `Async`.

Anyway, just my 2c :)

davesmith00000:
I'm inclined to say that we're coming at this from different angles, but it's also entirely possible that I'm just not following how or why one would want to do that - or possibly just how it's supposed to work. 😅 

Using an `F[A]` says nothing about how the sub will be run or its lifecycle. Through one lens that could be seen as a good thing: Why leak the implementation detail? And why impose so much structure? However, all subscriptions ever are is long running JS callbacks (a one off side effect is a `Cmd`). I don't want to hide the JavaScript from people (Elm's mistake IMO), I kinda want them to understand what they're doing (if they're going to write their own subs - probably hardly ever since you can do 90% of the things you'd like to do with `Sub.fromEvent`), which is registering a callback `Either[Throwable, A] => Unit` that can optionally be removed `F[Option[F[Unit]]]`, and the value optionally turned into a message `A => Option[Msg]`.

I will review your comment again tomorrow and read around the docs some more in case I realise I've got the wrong end of the stick. (Wouldn't be the first time... 😉)

armanbilge:
Ah, yes it may a thing of angles :)

> Using an `F[A]` says nothing about how the sub will be run or its lifecycle.

This is a really interesting comment :) I was reading this blog post yesterday, which I think said something very similar.

https://alexn.org/blog/2022/04/18/scala-oop-design-sample/

Specifically, this part:
> We say `F[_]` in this API, but as its authors, we know that we are talking of a data type that encapsulates / suspends side effects, but without saying so. This is the opposite of parametricity, and in this context usage of this style is not good API design, unless you really, really want tagless final in your project, and thus rely on conventions (e.g. whenever you see `F[_]`, you see `IO`). The danger being that readers of this code might not get a good mental model of using this API, because there’s information missing, and that’s bad.

> To be clear — **this isn’t good API design because `F[_]` is left unspecified.**

The words in that post are still sinking in for me. But at the surface, I totally get how just `F[A]` in an API that basically says nothing, which seems to mirror Alexandru's point about missing information.

Thanks for the insightful reply (also the comments about hiding JS are interesting), it's given me a lot to think about as well.

tyrian/js/src/main/scala/tyrian/Sub.scala
```diff
@@ -20,131 +23,181 @@ import scala.scalajs.js
...
+    Sub.make[F, js.Date, Int](id) { callback =>
+      Sync[F].delay {
+        dom.window.setInterval(
+          Functions.fun0(() => callback(Right(new js.Date()))),
+          interval.toMillis.toDouble
+        )
+      }
+    } { handle =>
+      Sync[F].delay(dom.window.clearTimeout(handle))
+    }
```
armanbilge, line 183: Side note: if possible, I'd still consider using `Temporal` here instead of `Sync`. I have no idea if it's important for frontend, but the Cats Effect test runtime is capable of mocking time, but only if you use the CE clock/time primitives rather than the raw APIs directly. For example, this lets you run tests with delays and stuff without actually having to wait for that time to pass, since it can instantly advance the clock, which is pretty cool.

https://typelevel.org/cats-effect/docs/core/test-runtime#mocking-time

---

Also, while I'm here I realize that I totally goofed the snippet I gave before, when I said this you could express `every` like this:
```scala
Temporal[F].sleep(interval) >> Temporal[F].realTimeDate
```
This would only run once and not _every_ interval 🤦 sorry about that. You could wrap a `repeatForever` around that, but then all that stuff about `join`ing the fiber is all wrong.

Without `Stream` I guess you _do_ need some sort of callback here. So, now I appreciate your design a lot better :)

davesmith00000: I had fun nosing around `Temporal` for a while there, but you're right, we're back to needing streams.

Interesting thing here is that you're unlikely to test that. Someone creates a `Sub` that emits a `Msg` after every elapsed interval. The `Msg` is handled in the `update` function. You'd test the update function, but you wouldn't bother testing the `Sub.every` - there's nothing to go wrong.

Of course, if someone wanted to do something fancy in their own custom `Sub`, they could certainly use a `Temporal` instance there if it helped them.

So at least for now, I think I'll leave it as-is.

tyrian/js/src/main/scala/tyrian/Cmd.scala
```diff
@@ -1,67 +1,63 @@
...
+sealed trait Cmd[+F[_], +Msg]:
+  /** Transforms the type of messages produced by the command */
+  def map[OtherMsg](f: Msg => OtherMsg): Cmd[F, OtherMsg]
+
+object Cmd:
+  given CanEqual[Cmd[_, _], Cmd[_, _]] = CanEqual.derived
+
+  extension [F[_], Msg, LubMsg >: Msg](cmd: Cmd[F, Msg])
+    def combine(other: Cmd[F, LubMsg]): Cmd[F, LubMsg] = Cmd.merge(cmd, other)
+    def |+|(other: Cmd[F, LubMsg]): Cmd[F, LubMsg]     = Cmd.merge(cmd, other)
```
armanbilge, line 18: I think you can put these directly on `Cmd` by moving the `LubMsg` parameter to the method. I think this helps API discoverability for users and is easier on the compiler as well.

Also I guess this is a Monoid? That's pretty cool :)

davesmith00000: I was struggling with that, it used to live on `Cmd`. It wasn't the `LubMsg` that was causing me problems, it was the covariant `F`. Help most welcome.

Yes, `Sub` and `Cmd` are monoids. I love monoids. You should see Indigo - it's monoid-tastic! 😁 

Originally there were cats monoid instances of these available for use, but I removed them when I removed cats. Could bring them back now if there was any value in it.

armanbilge: Ah, right, you need to do the exact same thing for `F[_]` too :) e.g. see the signatures in fs2:
```scala
  def concurrently[F2[x] >: F[x], O2](
      that: Stream[F2, O2]
  )(implicit F: Concurrent[F2]): Stream[F2, O] = {
```

davesmith00000: Works like a charm! Have pushed the change.

I did try that -except- that I didn't know about the little trick with the `x` you've got there. If you just use `_` you get an error: `unreducible application of higher-kinded type F to wildcard arguments`

Thank you kindly. 😄

davesmith00000: Monoid instances are also back. I even added tests.

armanbilge (approved):
Hopefully not premature 😆 but this has to land sometime, there's always opportunity for follow-up iteration, and overall big 👍

Truthfully I haven't reviewed most of the changes here, since I focused mostly on `Cmd` and `Sub`. But they seem important and they look really good :) and I'm happy to be pointed towards anything else to take a look at.

I suspect `Sub` and `fs2.Stream` are exploring a similar design space (effects that can emit more than once), so it makes a lot of sense to me that they should remain independent and that in fact `fs2.Stream` does not belong in Tyrian's public API at all.

Additionally, the raw callbacks in `Sub` that I had a hard time wrapping my head around at first: I've also come to appreciate that design :) it is very much like a "streaming" analog to CE's `Async[F].async` unsafe-interop method. This is an extremely common pattern in JS, and while it is possible to implement it using `Dispatcher`s and `Queue`s and `Stream`s ... it is _very_ nice to just be able to do so directly!

tyrian/js/src/main/scala/tyrian/Cmd.scala
```diff
@@ -1,67 +1,76 @@
...
+    def apply[F[_], Msg](cmds: Cmd[F, Msg]*): Batch[F, Msg] =
       Batch(cmds.toList)
+
+  given [F[_], Msg]: Monoid[Cmd[F, Msg]] = new Monoid[Cmd[F, Msg]] {
+    def empty: Cmd[F, Msg]                                   = Cmd.Empty
+    def combine(a: Cmd[F, Msg], b: Cmd[F, Msg]): Cmd[F, Msg] = Cmd.merge(a, b)
+  }
+
+  def combineAll[F[_], A](list: List[Cmd[F, A]])(implicit M: Monoid[Cmd[F, A]]): Cmd[F, A] =
+    list.foldRight(M.empty)(M.combine)
```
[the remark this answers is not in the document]
davesmith00000: Interesting: I merged it, and went in to apply the same fix to `Sub`s (2 minute lunchtime job) - but this actually breaks the tests. Looks like `Monoid[A].combineAll` does a `foldLeft` rather than a `foldRight`. Which means that instead of this:

```scala
Cmd.Combine(
  Cmd.Emit(10),
  Cmd.Combine(
    Cmd.Emit(20),
    Cmd.Emit(30)
  )
)
```

You get this:

```scala
Cmd.Combine(
  Cmd.Combine(
    Cmd.Emit(10),
    Cmd.Emit(20)
  ),
  Cmd.Emit(30)
)
```

It amounts to the same thing in the end, but less of a harmless refactor than I anticipated.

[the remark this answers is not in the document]
davesmith00000: Fixed and pushed - thanks!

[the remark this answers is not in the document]
armanbilge: Right, those have to be the same thing or else it violates the Monoid associativity law  `(a |+| b) |+| c == a |+| (b |+| c)`.

Running Cats discipline tests would also catch this :)

tyrian/js/src/main/scala/tyrian/Navigation.scala
```diff
@@ -1,42 +1,48 @@
...
+import cats.effect.kernel.Async
 import org.scalajs.dom.window
 import tyrian.Sub
 import tyrian.facades.HashChangeEvent
 
 import scala.scalajs.js
 import scala.util.control.NonFatal
 
+/** Provides simple routing based on url hash (anchor), such as: `http://mysite.com/#page1` */
 object Navigation:
```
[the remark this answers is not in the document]
davesmith00000: We briefly discussed this one privately, but just for the record and any curious readers: Tagless Final support isn't really what I'm after here (I don't think).

Here's an example of a few of those things you mentioned in use as the body of the `init` function:

```scala
val cmds: Cmd[IO, Msg] =
  Cmd.Batch(
    HotReload.bootstrap(hotReloadKey, Model.decode) {
      case Left(msg)    => Msg.Log("Error during hot-reload!: " + msg)
      case Right(model) => Msg.OverwriteModel(model)
    },
    Logger.info(flags.toString),
    Navigation.getLocationHash {
      case Navigation.Result.CurrentHash(hash) => Msg.NavigateTo(Page.fromString(hash))
      case _                                   => Msg.NavigateTo(Page.Page1)
    }
  )

(Model.init, cmds)
```

The thing about the TEA pattern is that you do not do all of your work within some `F` type, as you'd expect to do if you were building a Scala microservice. Most of your code is pretty vanilla 'unsafe' functions that could do anything, and right at the edges (as in the example above) you can describe effects. Each of the main functions that make up a TEA pattern program are as pure as you code them to be. Those edge effects are processed in the background for you by the runtime. As such, an `F` with capabilities would not be passed around, it does not pervade the system, it's only used locally in the `Cmd`/`Sub` implementation itself (and sure, you can use a low power typeclass there by all means). All the end users are going to see is a concrete `IO`.

This would not be true if I changed the API to something sort of like the below, where everything is calculated within an abstract `F` type:

```diff
- def update(msg: Msg, model: Model): (Model, Cmd[IO, Msg]) = ???
+ def update(msg: Msg, model: Model): F[(Model, Cmd[IO, Msg])] = ???
```

People can agree to disagree on whether the TEA pattern is a good idea or not, but that is the pattern. I suspect FRP might be able to cope with Streams and Tagless Final, but it feels a bit much here.

The reason I like the TEA pattern and Tyrian's flavour of it in particular is that while it isn't the most correct and pure form of Scala FP out there, it is very approachable, easy to use, and easy to achieve things in. I'd hate to spoil that for the users with too much in the way of heavy FP. I can imagine that some (I don't mean you) might think that's a bit of an anti-intellectual idea, but I think it's just a design choice. There's something very clean and straightforward about the APIs that I really enjoy.

davesmith00000:
mario

Mario is here to say that testing looks good. I think we're done here.

PR #91, opened 2022-04-21
Swap `dom.HashChangeEvent` for internal facade

armanbilge:
Noticed this, so I was going to upstream it into scala-js-dom ... but then I discovered that it was already there :)


davesmith00000 (approved):
LGTM!

PR #92, opened 2022-04-21
Fixed #90: Rename Sub/Cmd Empty to None

### LINKED ISSUE #90: Rename Cmd/Sub Empty to None
Empty is a weird term, and I notice Elm uses None there instead which makes more sense.

davesmith00000:


hobnob (approved):
👍

PR #94, opened 2022-04-21
Fixed #87: Update to M => Msg => (M, Cmd[F, Msg])

### LINKED ISSUE #87: Change update to `Model => Msg => (Model, Cmd[F, Msg])`
The only thing you ever do in `update` is match on the Msg type... so a better formulation than the current:

```scala
def update(msg: Msg, model: Model): (Model, Cmd[F, Msg])
```

would be

```scala
def update(model: Model): Msg => (Model, Cmd[F, Msg])
```

davesmith00000:


davesmith00000:
@hobnob, what do you think of this?

hobnob (approved):
Looks good :-)

hobnob:
Generally looks good - but the build is failing in CI... so long as you're happy that's not a problem, I'm happy :-)

PR #99, opened 2022-05-06
Conversion to use scala-js-snabbdom

davesmith00000:
This is a WIP PR that replaces snabbdom with [scala-js-snabbdom](https://github.com/buntec/scala-js-snabbdom).

Currently it doesn't quite work - very close - but not quite.


tyrian/js/src/main/scala/tyrian/runtime/TyrianRuntime.scala
```diff
@@ -139,36 +138,51 @@ final class TyrianRuntime[F[_]: Async, Model, Msg](
...
+            (n, (e: dom.Event) => onMsg(msg.asInstanceOf[dom.Event => Msg](e)))
           }
 
-        h(name, obj(props = props, attrs = as, on = events))(childrenElem: _*)
+        val data: VNodeData =
+          VNodeData.builder
+            .withProps(props: _*)
+            .withAttrs(as: _*)
+            .withOn(events: _*)
+            .build
```
davesmith00000, line 160: This I don't like. I can't provide nice collections, and I have to use this funky builder thing. I want to just do:
```scala
VNodeData.empty.withProps(props)
```
Where props is a `List`|`Array`|`js.Array`|`Whatever`.

tyrian/js/src/main/scala/tyrian/runtime/TyrianRuntime.scala
```diff
@@ -139,36 +138,51 @@ final class TyrianRuntime[F[_]: Async, Model, Msg](
...
+        EventListeners.module,
+        Dataset.module
+      )
     )
 
   def render(oldNode: Element | VNode, model: Model): VNode =
-    patch(oldNode, toVNode(view(model)))
+    oldNode match
+      case em: Element => patch(em, toVNode(view(model)))
+      case vn: VNode   => patch(vn, toVNode(view(model)))
```
davesmith00000, line 185: Would be nice is patch accepted `Element | VNode`, since they are both valid.

tyrian/js/src/main/scala/tyrian/runtime/TyrianRuntime.scala
```diff
@@ -139,36 +138,51 @@ final class TyrianRuntime[F[_]: Async, Model, Msg](
...
-
-        val props =
-          js.Dictionary(attrs.collect { case Property(n, v) => (n, v) }: _*)
+          }
 
-        val events =
-          js.Dictionary(attrs.collect { case Event(n, msg) =>
-            (n, fun((e: dom.Event) => onMsg(msg.asInstanceOf[dom.Event => Msg](e))))
-          }: _*)
+        val props: List[(String, PropValue)] =
```
davesmith00000, line 147: `PropValue` is an alias to.. `Any`? I realise props can be numbers or strings or booleans, but in HTML land they're all strings in fact? Because they're written into attributes? Might be fine, feels odd.

davesmith00000:
FAO: @buntec and @armanbilge

Nearly! It nearly works! Well done! 👏

tyrian/js/src/main/scala/tyrian/runtime/TyrianRuntime.scala
davesmith00000: I think the thing here is: Someone mad enough to use this lib directly would benefit from the set up you have here. If you're building a library on top of it though, it's just inefficient. Though to be fair if efficiency is my concern, I just probably just use the main constructor or do a single `VnodeData.empty` and... set the `vars`?

tyrian/js/src/main/scala/tyrian/runtime/TyrianRuntime.scala
armanbilge: No, I don't think so actually. Props and attributes are different. Of course parseable HTML can only be expressed with attributes, but from JS APIs you'd set the props as their native values.

tyrian/js/src/main/scala/tyrian/runtime/TyrianRuntime.scala
armanbilge: I agree with you on this one. At the very least it's worth offering the alternate constructor.

> Though to be fair if efficiency is my concern, I just probably just use the main constructor or do a single `VnodeData.empty` and... set the `vars`?

I'd be shocked if the Scala.js optimizer doesn't inline this to the same thing. Actually, we can make this explicit by adding `@inline` annotations. It's a mutable builder so it's just friendly syntax for setting the `vars` on a `VNodeData.empty`.

tyrian/js/src/main/scala/tyrian/runtime/TyrianRuntime.scala
davesmith00000: Yep fair enough, I think I was just a bit taken aback to see the Any, but it may be unavoidable.

tyrian/js/src/main/scala/tyrian/runtime/TyrianRuntime.scala
```diff
@@ -139,36 +138,51 @@ final class TyrianRuntime[F[_]: Async, Model, Msg](
...
+          attrs.collect { case Property(n, v) => (n, v) }
 
-        val childrenElem: List[VNodeParam] =
-          children.map {
-            case t: Text            => VNodeParam.Text(t.value)
-            case subHtml: Html[Msg] => VNodeParam.Node(toVNode(subHtml))
+        val events: List[(String, dom.Event => Unit)] =
+          attrs.collect { case Event(n, msg) =>
+            (n, (e: dom.Event) => onMsg(msg.asInstanceOf[dom.Event => Msg](e)))
           }
```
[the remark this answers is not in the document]
armanbilge: Ouch, that's really weird. I wonder if we haven't ported a test covering this yet, or it's just not covered by the snabbdom suite.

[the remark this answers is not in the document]
davesmith00000: Or I'm not using it right. 😅 
Mostly this is a blind port of the original implementation I inherited. I'm always prepared to accept that this might be me. 😉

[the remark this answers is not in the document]
armanbilge: We are probably just missing the tests :) I'd guess they're covered by this suite which we have yet to port.
https://github.com/snabbdom/snabbdom/blob/fed2f31b8990258839e4547b967d3d9f21a94032/test/unit/eventlisteners.ts

build.sbt
```diff
@@ -93,9 +93,10 @@ lazy val tyrian =
     .jsSettings(
       commonJsSettings,
       libraryDependencies ++= Seq(
-        "org.scala-js"  %%% "scalajs-dom"        % Dependancies.scalajsDomVersion,
-        "org.typelevel" %%% "cats-effect-kernel" % Dependancies.catsEffect,
-        "co.fs2"        %%% "fs2-core"           % Dependancies.fs2
+        "org.scala-js"      %%% "scalajs-dom"        % Dependancies.scalajsDomVersion,
+        "org.typelevel"     %%% "cats-effect-kernel" % Dependancies.catsEffect,
+        "co.fs2"            %%% "fs2-core"           % Dependancies.fs2,
+        "com.github.buntec" %%% "scala-js-snabbdom"  % "0.0-29f78bf-SNAPSHOT"
```
armanbilge, line 99:
```suggestion
        "io.github.buntec" %%% "scala-js-snabbdom"  % "0.0-1c017a3-SNAPSHOT"
```

You'll also need to add:
```scala
resolvers +=
  "Sonatype S01 OSS Snapshots" at "https://s01.oss.sonatype.org/content/repositories/snapshots"
```

tyrian/js/src/main/scala/tyrian/runtime/TyrianRuntime.scala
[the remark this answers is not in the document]
armanbilge: Hopefully should be fixed as of https://github.com/buntec/scala-js-snabbdom/pull/15.

build.sbt
davesmith00000: That is not [published](https://s01.oss.sonatype.org/content/repositories/snapshots/com/github/) it seems, but I'll try a local publish again.

armanbilge: The group id changed `com.github.buntec` -> `io.github.buntec`.

https://s01.oss.sonatype.org/content/repositories/snapshots/io/github/buntec/scala-js-snabbdom_sjs1_2.13/0.0-1c017a3-SNAPSHOT/

davesmith00000: Sorry - my bad - you changed the group to `io`...

davesmith00000: Updated, that does indeed appear to fix the issues. 😄

build.sbt
```diff
@@ -93,9 +95,10 @@ lazy val tyrian =
     .jsSettings(
       commonJsSettings,
       libraryDependencies ++= Seq(
-        "org.scala-js"  %%% "scalajs-dom"        % Dependancies.scalajsDomVersion,
-        "org.typelevel" %%% "cats-effect-kernel" % Dependancies.catsEffect,
-        "co.fs2"        %%% "fs2-core"           % Dependancies.fs2
+        "org.scala-js"     %%% "scalajs-dom"        % Dependancies.scalajsDomVersion,
+        "org.typelevel"    %%% "cats-effect-kernel" % Dependancies.catsEffect,
+        "co.fs2"           %%% "fs2-core"           % Dependancies.fs2,
+        "io.github.buntec" %%% "scala-js-snabbdom"  % "0.0-1c017a3-SNAPSHOT"
```
armanbilge, line 101:
```suggestion
        "io.github.buntec" %%% "scala-js-snabbdom"  % "0.1.0-M2"
```

https://repo1.maven.org/maven2/io/github/buntec/scala-js-snabbdom_sjs1_3/0.1.0-M2/

davesmith00000:
The build is failing, but it's failing for the same reason `main` branch is failing, and that will be resolved via the other open PR.

Have decided to be brave and release Tyrian against `scala-js-snabbdom` `0.1.0-M2`.

What could possibly go wrong? 😁

PR #109, opened 2022-06-06
Fix Body constructors

chuwy:
Presumably it should be working as:

```scala
Body.json("{}")
```

whereas currently it works only as:

```scala
Body.Empty.json("{}")
```

No tests seem to be affected.


davesmith00000:
Well spotted! Will merge soon. Thanks. 😄

davesmith00000 (approved):
👍

PR #111, opened 2022-06-06
Potential bug: add an example reproducing a problem with IO.sleep

chuwy:
I'm running into a problem, where I have `Model` with `List[Notification]` and trying to show the notifications in OS-like popups. The notifications have quite naive approach:

```scala
val sleep = Cmd.SideEffect(IO.sleep(5.seconds))
val clean = Cmd.Emit(Msg.Remove(msg)))
sleep.combine(clean)
```

Basically, I'm adding a notification, then setting a timer to remove it. In my app notifications appear with a big delay and in batches rather than one by one. In the minimal example they do appear without problems, but I've also noticed they disappear in batches and after a period much longer than expected 5 seconds.

Wondering if there's anything wrong with the approach I'm using.


chuwy:
For the record, there are three loosely coupled problems:

1. I used `Cmd.combine` which is wrong for what I wanted to achieve as RHS command gets executed immediately. My fault
2. Actual bug (#112)
3. Another actual bug (#113)

davesmith00000:
This PR was super helpful debugging the problem - thanks! 😄 
I shall close this PR without merging. If you think there's value in having this example around anyway, let me know and we can chat about it.

PR #116, opened 2022-06-07
Need for accurate speed

davesmith00000:
This PR revisits the runtime implementation to do four things:

1. Decrease time to first page draw
2. Run `Cmd`s in parallel
3. Evaluate tasks in the order sent
4. Split update from render

### DIFF
--- a/tyrian/js/src/main/scala/tyrian/TyrianAppF.scala
+++ b/tyrian/js/src/main/scala/tyrian/TyrianAppF.scala
@@ -12,7 +12,10 @@ import scala.scalajs.js.annotation._
   */
 trait TyrianAppF[F[_]: Async, Msg, Model]:
 
-  def MaxConcurrentMessages: Int = 1024
+  /** Specifies the number of queued tasks that can be consumed at any one time. Default is 1024 which is assumed to be
+    * more than sufficient, however the value can be tweaked in your app by overriding this value.
+    */
+  def MaxConcurrentTasks: Int = 1024
 
   val run: Resource[F, TyrianRuntime[F, Model, Msg]] => Unit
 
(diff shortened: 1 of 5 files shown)


davesmith00000:
image

davesmith00000:
(cc @chuwy)

davesmith00000:
I have no way to demonstrate this, but the rendering is so much smoother in this version.

tyrian/js/src/main/scala/tyrian/TyrianAppF.scala
```diff
@@ -12,6 +12,8 @@ import scala.scalajs.js.annotation._
   */
 trait TyrianAppF[F[_]: Async, Msg, Model]:
 
+  def MaxConcurrentMessages: Int = 1024
```
chuwy, line 15: Curious as of how to tweak this value for a real-world appliaction.

chuwy (dismissed):
Looks amazing!

tyrian/js/src/main/scala/tyrian/TyrianAppF.scala
davesmith00000: Yeah I was wondering about that. If you know about it you can just override it, but I don't normally do overrides. Indigo requires you to provide a config object as part of the `init` function, I could do that again even though it's just one field (for now).

In practice 1024 is _massive_. I'd expect most apps to rarely even produce 10 at a time.

davesmith00000: Decided to leave it for now, but I've renamed it to `MaxConcurrentTasks` and added scaladocs explaining that it can be overridden. We'll see how it behaves in the wild and revisit as necessary.

PR #117, opened 2022-06-07
Fixed #115: Cmd.Emit can be delayed

### LINKED ISSUE #115: Add a CMD like Emit with a delay

davesmith00000:
(cc @chuwy)

### DIFF
--- a/tyrian/js/src/main/scala/tyrian/Cmd.scala
+++ b/tyrian/js/src/main/scala/tyrian/Cmd.scala
@@ -4,6 +4,8 @@ import cats.effect.kernel.Sync
 import cats.kernel.Monoid
 
 import scala.annotation.targetName
+import scala.concurrent.duration.FiniteDuration
+import scala.concurrent.duration.*
 
 /** A command describes some side-effect to perform.
   */
@@ -30,6 +32,9 @@ object Cmd:
       case (c1, c2)             => Cmd.Combine[F, LubMsg](c1, c2)
     }
 
+  def emitAfterDelay[Msg](msg: Msg, delay: FiniteDuration): Emit[Msg] =
+    Emit(msg, delay)
+
   /** The empty command represents the absence of any command to perform */
   case object None extends Cmd[Nothing, Nothing]:
     def map[OtherMsg](f: Nothing => OtherMsg): None.type = this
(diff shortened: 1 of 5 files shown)


tyrian/js/src/main/scala/tyrian/Cmd.scala
```diff
@@ -42,8 +47,12 @@ object Cmd:
       SideEffect(Sync[F].delay(thunk))
 
   /** Simply produces a message that will then be actioned. */
-  final case class Emit[Msg](msg: Msg) extends Cmd[Nothing, Msg]:
-    def map[OtherMsg](f: Msg => OtherMsg): Emit[OtherMsg] = Emit(f(msg))
+  final case class Emit[Msg](msg: Msg, afterDelay: FiniteDuration) extends Cmd[Nothing, Msg]:
```
chuwy, line 50: Sorry for being noisy, but I still think that if we can express something with a very simple combination of existing commands (`Run(IO.sleep(delay), _ => msg)`) - we should do that rather than overloading a single-purpose command.

So, I'd prefer to keep the `Emit` as is and changed `emitAfterDelay` to:

```scala
def emitAfterDelay[Msg](msg: Msg, delay: FiniteDuration): Emit[Msg] =
  Run(IO.sleep(delay), _ => msg)
```

davesmith00000:
I'm still pondering your other comment about keeping `Emit` as it was. I don't object strongly to it since it's the same sort of pattern that `Sub` follows, so we have precedent. Equally, under the covered Emit is transforming into an `IO` (or whatever), so decorating it slightly doesn't offend me.

Philosophically, I lean towards easy of use / helpful / friendly apis / low barrier to entry over 'FP correct'. Cats Effect for example is incredibly composable and brilliant to use, and also _really, really_ hard to learn in the first place. Some feel that a library being super composable makes it friendly to use, but I don't share that opinion (or rather, my views on it aren't not so binary). I'm not criticising CE there, just using it as a recognisable example.

What about this for running sequential effects with msg/no-msg? Works right now! Batch could be a `NEL`, but it's handy to be able to do things like `if p then Batch(cmd) else Batch()`.

```scala
Cmd.Batch(
  Cmd.SideEffect(task1), // No msg
  Cmd.Run(task2, toMsg) // produces msg.
)
```

What that does not let you do is pipe the result of one into the next, but I don't think you're asking for that, of if you are then the options currently are either (depending what you're doing):

1. `IO`s in a for-comp, wrapped in a `Cmd.Run`. That doesn't help with the built in Http client so...
2. Your comment above is correct:

> Technically, I can just catch Msg.GotResponse in my update and emit Msg.Notify from there. Which is probably the right thing to do.

Separately, you're getting caught out by `Combine`. Combine is taken from the original Elm implementation and I'm trying to stay close-ish to Elm. It's a Monoid append on `Cmd` based on the fact that `List` is a monoid. So it's 'one unrelated task then the next' not a monad-like sequencing operation. If that helps.

chuwy:
Ok, `Batch` is a great suggestion, let's ignore that NEL part.

My last argument in defense of current ADT is that we can consider ADT itself a more lower-level, internal API and provide end-users with many handy combinators like `emitAfterDelay`. Like what circe does for `Json` or ZIO and CE with their tons of smart constructors.

Extending `Cmd` with things that we easily can do seems like a dangerous precendent opening a room for less sensible UX-improvements like:

```scala
final case class GetTime[F[_]]() extends Cmd[F, FiniteDuration]:
```

Having said that, I certainly agree it's just a matter of personal preference.

davesmith00000:
Closing this PR.

I've boldly gone right ahead and [committed the alternative](https://github.com/PurpleKingdomGames/tyrian/commit/190ff40fe6efd12a6a83539306fa06c0949b3a28) to `main`, but I think you'll approve @chuwy (hopefully). 😉

PR #120, opened 2022-06-11
Http.send: fix setting headers

chuwy:
It appears that we must set headers before calling `open`. Othersise I'm getting an exception after any request with `withHeaders`.

More details: https://stackoverflow.com/questions/28524485/xmlhttprequest-throwing-invalidsateerror-saying-object-state-must-be-opened


davesmith00000 (approved):
Thanks!

Issue #130, opened 2022-08-07
Publish Tyrian with ZIO support


davesmith00000:
I'm setting up this issue just to track progress, as I've been trying to organise this since I did the initial move across to Cats Effect 3.

At first the [issue](https://github.com/zio/interop-cats/issues/528) was that the libraries hadn't been published for Scala 3 + Scala.js + CE3. That appears to have been resolved naturally, so I've closed the ticket.

The next problem is that the versions of CE and fs2 that the interop lib is published against is [way behind](https://github.com/zio/interop-cats/issues/596). It does look like the codebase itself has been brought up to date though, so hopefully we're just waiting on a new publish.

After that it's just a matter of making it work. There is [an example in the README](https://github.com/zio/interop-cats#async) of what is needed (more or less).

davesmith00000:
This is a WIP PR for that: https://github.com/PurpleKingdomGames/tyrian/pull/134

davesmith00000:
Will be in the next release.

PR #134, opened 2022-08-16
Add support for ZIO 2.0

davesmith00000:
I believe this is blocked from going any further yet by missing dependencies.

https://github.com/zio/interop-cats/issues/596

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -135,7 +135,11 @@ lazy val tyrianZIO =
     .in(file("tyrian-zio"))
     .settings(
       name := "tyrian-zio",
-      commonSettings ++ publishSettings
+      commonSettings ++ publishSettings,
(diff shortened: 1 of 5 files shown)


davesmith00000:
Relates to: https://github.com/PurpleKingdomGames/tyrian/issues/130

davesmith00000:
This all works locally but is dependant on:
https://github.com/zio/interop-cats/pull/603
https://github.com/zio/interop-cats/issues/596
...and on a release.

davesmith00000:
The merge/release of this is no longer blocked as Tyrian itself does not now depend on the interop-cats lib. Users have to bring in the lib at the point of use, but that's acceptable. (Thanks to @armanbilge for the suggestion.)

PR #136, opened 2022-08-16
Improve launch behaviour

### LINKED ISSUE #135: Allow `launch` to be called with an element, not just a container ID
Follow on from: https://github.com/PurpleKingdomGames/tyrian/issues/131

Relevant description:

> Could launch accept the DOM element in addition to containerId?
> I imagine not everyone wants to use id, or already have a reference to the DOM node and wants to use it instead.

davesmith00000:


davesmith00000:
(cc @grzegorz-bielski)

tyrian/js/src/main/scala/tyrian/TyrianAppF.scala
```diff
@@ -42,29 +43,56 @@ trait TyrianAppF[F[_]: Async, Msg, Model]:
...
         subscriptions,
         MaxConcurrentTasks
       )
     )
+
+  @SuppressWarnings(Array("scalafix:DisableSyntax.throw"))
+  private def runReadyOrError(containerId: String, flags: Map[String, String]): Unit =
+    Option(document.getElementById(containerId)) match
+      case Some(e) =>
+        ready(e, Map[String, String]())
```
grzegorz-bielski, line 95: Don't we want to pass `flags` to `ready` ? 👀

davesmith00000: Ooops! Thank you. I guess that's why I code in the evening and review in the morning. 😄 
Will fix this evening before I merge it in. I do miss Scala 2's 'unused' compiler warnings at times like this...

PR #140, opened 2022-08-28
Add helper to make `Sub` from `fs2.Stream`

armanbilge:


davesmith00000:
I've done a basic test of this and it does indeed work. I haven't tested removing a running sub yet though... that's my next trick.

```scala
  def stream: Stream[IO, Msg] =
    Stream.awakeEvery[IO](5.seconds)
      .map(t => Msg.Log("Log after " + t.toString))

  def sub: Sub[IO, Msg] =
    Sub.make("pulse", stream)
```

Produces:

image

davesmith00000:
Also tested removing and re-adding, all seems to work perfectly. Nice one @armanbilge. 😄

davesmith00000 (approved):
Short and sweet!

PR #144, opened 2022-09-10
Fixed #143: Extension syntax to make Cmds/Subs

### LINKED ISSUE #143: Extension methods to convert any Async F of A into a Cmd or Sub
Also any `fs2.Stream` into a Sub. Currently you have to wrap this stuff up in something like a `Cmd.Run` but it might be nice to let people skip that if they like.

davesmith00000:
Basically for certain types of value you to create a `Cmd` or `Sub` by adding `toCmd` or `toSub` to the end.

It isn't complicated, but it's kinda fun because you can just for-comp a nice task based process together and then say `myFunkyIOThing.toCmd` at the end.

Added examples to the Sandbox just to prove to myself they worked.

More advanced extension methods get complicated quickly so I'm stopping here. I tried to do versions that would convert lists to `Cmd`s to a `Cmd.Batch` but the type inference fell to pieces. Creating `Sub`s is also not straight-forward in general, in fact the `fs2.Stream` version we have here is by far the simplest approach. After that you're better off with one of the existing `Sub` constructors so that you get some clue that you need to manufacture a callback.

### DIFF
--- a/tyrian/js/src/main/scala/tyrian/syntax.scala
+++ b/tyrian/js/src/main/scala/tyrian/syntax.scala
@@ -12,7 +12,7 @@ object syntax:
 
   /** Make a cmd from any `F[A]`
     */
-  extension [F[_], A, Msg](task: F[A])
+  extension [F[_], A](task: F[A])
     def toCmd: Cmd.Run[F, A, A] =
       Cmd.Run(task)(identity)
 
--- a/tyrian/js/src/main/scala/tyrian/syntax.scala
+++ b/tyrian/js/src/main/scala/tyrian/syntax.scala
@@ -1,0 +1,23 @@
+package tyrian
+
+import cats.effect.Async
+
+object syntax:
+
+  /** Make a side effect cmd from any `F[Unit]`
+    */
+  extension [F[_]](task: F[Unit])
+    def toCmd: Cmd.SideEffect[F] =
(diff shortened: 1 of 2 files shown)


tyrian/js/src/main/scala/tyrian/syntax.scala
```diff
@@ -0,0 +1,23 @@
...
+    */
+  extension [F[_]](task: F[Unit])
+    def toCmd: Cmd.SideEffect[F] =
+      Cmd.SideEffect(task)
+
+  /** Make a cmd from any `F[A]`
+    */
+  extension [F[_], A, Msg](task: F[A])
+    def toCmd: Cmd.Run[F, A, A] =
+      Cmd.Run(task)(identity)
```
armanbilge, line 17: `Msg` doesn't seem to be used here :)

davesmith00000: Thank you. 🙂
This is why I make PR's even when I think I'm the only person who will look at them. Fresh eyes in the morning.

davesmith00000: Yes, at one point there was another method in here next to this one that used `Msg` like:

`def toCmd(using toMsg: A => Msg): Cmd.Run[F, A, Msg]`

But it was more confusing that useful.

PR #150, opened 2022-09-28
Remove use of NonFatal exception

davesmith00000:
Noticed while investigating issues with Subs.

Adding `NonFatal` as the catch condition on the Navigation Sub was well intentioned, but the result is that we're leaking Java style errors in a JavaScript environment, which was the intended behaviour. In this case we're doing an `indexOf` inside a `try` block with the full expectation of it being caught, but it isn't. JS throws the exception and carries on regardless.. so what's the point of the exception or the try?

### DIFF
--- a/sandbox/src/main/scala/example/Sandbox.scala
+++ b/sandbox/src/main/scala/example/Sandbox.scala
@@ -515,7 +515,7 @@ object Page:
       case "page5"  => Page5
       case "#page6" => Page6
       case "page6"  => Page6
-      case s        => Page1
+      case _        => Page1
 
 object Model:
   // val echoServer = "ws://ws.ifelse.io" // public echo server
--- a/tyrian/js/src/main/scala/tyrian/HotReload.scala
+++ b/tyrian/js/src/main/scala/tyrian/HotReload.scala
@@ -6,7 +6,6 @@ import util.Functions
 
 import scala.concurrent.duration.FiniteDuration
 import scala.scalajs.js
-import scala.util.control.NonFatal
 
 /** A very simple mechanism to allow automatic loading and saving of your applications model to local storage. Uses:
   *
--- a/tyrian/js/src/main/scala/tyrian/Navigation.scala
+++ b/tyrian/js/src/main/scala/tyrian/Navigation.scala
@@ -18,16 +18,19 @@ object Navigation:
   /** Subscribes to changes in the url hash and reports when they occur */
   def onLocationHashChange[F[_]: Async, Msg](resultToMessage: Result.HashChange => Msg): Sub[F, Msg] =
     Sub.fromEvent("hashchange", window) { e =>
-      try {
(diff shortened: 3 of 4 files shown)


davesmith00000:
Oh that's interesting. I'm guessing this is something CE is deciding depending on fast or full linking?

Also my point is that in fact it isn't fatal, to me at least. I can recover safely. Soooo... :shrug: :smile:

armanbilge:
>  I'm guessing this is something CE is deciding depending on fast or full linking?

No, nothing to do with CE 🤔 sorry, maybe I drove-by too fast 😂 would you mind adding some more context?

The key words that caught my eye were something about `NonFatal` leaking exceptions with code intended to throw an indexing-relating exception. And I have certainly encountered this before e.g. an index-out-of-bounds on the JVM is an undefined behaviour on JS which is a fatal exception. But this is Scala.js semantics.

The only bit that's relevant to CE is that the fiber runloop basically surrounds every `IO(...)`, `flatMap`, etc. with a `try` / `catch NonFatal(ex)`.

davesmith00000:
> The key words that caught my eye were something about `NonFatal` leaking exceptions with code intended to throw an indexing-relating exception. And I have certainly encountered this before e.g. an index-out-of-bounds on the JVM is an undefined behaviour on JS which is a fatal exception. But this is Scala.js semantics.

That is what is happening here. :+1:

armanbilge:
Right, so if your logic relies on an exception, and you are getting an undefined behavior exception, that means that there will be no exception in production and things might not work.

davesmith00000:
There you go again with your fancy talk of production, what are you talking about? 😄

What's setting what 'in production'? I feel like I'm going to kick myself when I realise what you're saying. You mean Node maybe?

armanbilge:
😂 oh sorry. Basically the difference between fastLink (development) and fullLink (production). This is the terminology Scala.js uses to distinguish between them.

https://www.scala-js.org/api/scalajs-library/latest/scala/scalajs/LinkingInfo$.html

armanbilge:
See also:
https://www.scala-js.org/doc/semantics.html#undefined-behaviors

> `UndefinedBehaviorErrors` are _fatal_ in the sense that they are not matched by case `NonFatal(e)` handlers. This makes sure that they always crash your program as early as possible, so that you can detect and fix the bug. It is _never_ OK to catch an `UndefinedBehaviorError` (other than in a testing framework), since that means your program will behave differently in `fullLinkJS` stage than in `fastLinkJS`.

davesmith00000:
Ah, excellent! I wasn't a million miles off with an earlier comment then. 👍

Thank you for the pointer, I'll re-read the linked references when I review this (probably at the weekend now) and decide how to proceed.

It does make me wonder: What does stopping a program in a browser even mean? and is that what JS devs really do? ...but that's a philosophical discussion for another day.

hobnob (approved):
Looks good :)

PR #156, opened 2023-01-11
Add tags with raw HTML

daniel-ciocirlan:
This PR adds virtual DOM tags that can render arbitrary HTML inside.

Useful for rendering dynamic HTML content that comes outside the Tyrian application, e.g. backend returns meaningful content like Markdown, which then needs to be turned to HTML.

cc @armanbilge @davesmith00000

### DIFF
--- a/tyrian/js/src/main/scala/tyrian/runtime/Rendering.scala
+++ b/tyrian/js/src/main/scala/tyrian/runtime/Rendering.scala
@@ -39,9 +39,13 @@ object Rendering:
 
   def toVNode[Msg](html: Html[Msg], onMsg: Msg => Unit): VNode =
     html match
-      case RawTag(name, attrs, innerHTML) =>
+      case RawTag(name, attrs, html) =>
         val data = buildNodeData(attrs, onMsg)
-        h(name, data, innerHTML)
+        val elm  = dom.document.createElement(name)
+        elm.innerHTML = html
+        val vNode = snabbdom.toVNode(elm)
+        vNode.data = data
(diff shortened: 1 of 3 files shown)


davesmith00000:
Looks good, and thank you very much for the work.

The only thing that may be missing is a a bit of API-consistent-niceness.

Generally you don't write `Tag` to make a tag, you write `div(...)` or `span(...)` or whatever, like this:
https://github.com/PurpleKingdomGames/tyrian/blob/main/tyrian/shared/src/main/scala/tyrian/Html.scala#L47

...and it's just a method that creates the `Tag` for you. Maybe we could add a `raw(...)` delegate? Or something? I have the same thing in Ultraviolet:
https://github.com/PurpleKingdomGames/ultraviolet/blob/main/ultraviolet/shared/src/main/scala/ultraviolet/syntax.scala#L58

davesmith00000 (approved):
Thanks, LGTM. :+1:

I've tested this and it works as described.

I think there's a follow up piece of work to do (I'll raise an issue in a moment) to work on the DX of this. I don't know exactly what change I'm looking for right now and so I'm happy to wrap this up and get it merged, but having to name to new containing element feels clunky. I think your idea (from our chat on Discord) of being able to do something like this would be good:

```scala
div(id := "foo).innerHtml("<p>Hello</p>")
```

Where `innerHtml` is just a method on `Tag`, I expect, but having had a very quick snoop around the code I'm not yet sure how that would work, and I don't want it to be a blocker.

Thanks again!

PR #160, opened 2023-01-13
Fixed #157: Nix shell work

### LINKED ISSUE #157: Investigate/Reminder: Nix flake setup doesn't appear to support my linux machine

davesmith00000:
@gvolpe - Any chance you can give this a quick look and tell me all the bad things I've done? :grin: 

I found that the dev shell didn't work on my machine, so I simplified the systems bit to something I've been using in other projects, I also unified the devshells because it seemed sensible to me (but I could be wrong!), and ditched support for the normal nix-shell feature.

On that last point, I noted your comment about Nix Flakes being experimental... but I can't see them going back on it and I'm happy for the few people likely to work on Tyrian with Nix to take the hit if they do change anything substantially. Fair enough?

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,29 +21,27 @@ sbt clean update compile test +publishLocal
 
 There is also a `build.sh` script in the root folder, but that also builds the docs and example projects (based on the currently released version).
 
-### Nix dev-shells
+### Nix dev-shell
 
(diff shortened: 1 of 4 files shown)


gvolpe (approved):
LGTM 👍🏽 :)

Issue #149, opened 2022-09-28
Web Socket subscription drops initial messages


gvolpe:
Hi @davesmith00000 👋🏽 

I've been debugging all day (going crazy really 😄) a subtle issue with the Web Socket subscriptions and I now have a lead, but unfortunately not a fix yet.

As far I can observe, once a connection is established, there seems to be a delay in registering the subscription and so messages that arrive immediately are never processed by Tyrian, but they can be seen in the browser's developer console.

![msg](https://i.imgur.com/8nOwYhx.png)

I think the last version where this was working as expected was 0.3.2, so shame on me, I haven't properly tested my application to find this issue sooner.

I tried to minimize the example using the Echo Server from the examples. For example, there is a [message](https://github.com/PurpleKingdomGames/tyrian/blob/main/examples/websocket/server/websocketserver.js#L11) saying "Connection made...", which doesn't get to Tyrian subs. 

Long shot, but do you recall if this message was being processed on versions 0.3.2 and lower?

On my application, I temporarily added a delay before sending the `Attached` message, but it's not a long-term solution.

------

Unrelated question: Is there a Matrix bridge for the Discord channel?

davesmith00000:
> Unrelated question: Is there a Matrix bridge for the Discord channel?

Given I only vaguely know what the means the answer is almost certainly, no, but happy to look into it?

davesmith00000:
WebSockets: Well that's a shame! Your description implies its been broken since we introduced Cats Effect / fs2. I will look into it - hopefully this evening - and report back.

gvolpe:
RE: Matrix bridge, I don't have much experience in setting it up either, but the Scala channel does it, so maybe they may be able to help? AFAIK the easiest way is via https://t2bot.io/discord/ but I haven't tested it, happy to help on the Matrix side!

gvolpe:
FWIW I've solved this on the application side by adding an "acknowledgement" mechanism to ensure the message is received, which I think it's nice regardless of the current issue, giving stronger guarantees.

Probably worth looking into the WS echo server example to compare it with `0.3.2`, but don't lose your head with it! 😄 

Happy to help with what I can :)

davesmith00000:
Evening @gvolpe :wave: 

Mixed success. :smile: 

**Discord:**

t2bot looks easy enough, but I have no idea what I'm doing with Matrix itself. I'll try and read up more tomorrow.

**WebSockets**

[...]

PR #163, opened 2023-01-14
Websocket drops initial message

davesmith00000:
This is a fix for https://github.com/PurpleKingdomGames/tyrian/issues/149, largely based on the solution @gvolpe came up with for the [FEDA trading app](https://github.com/gvolpe/trading/blob/main/modules/ws-client/src/main/scala/trading/client/WebSocket.scala#L102).

The difference between the two solutions is only that I've created a new stream sub constructor that allows you to include additional clean up tasks directly, meaning we can clear the event listeners when the stream is cancelled.

Seems to work just fine for me... any chance you could try it out @gvolpe? :sweat_smile:

### DIFF
--- a/tyrian/js/src/main/scala/tyrian/Sub.scala
+++ b/tyrian/js/src/main/scala/tyrian/Sub.scala
@@ -121,10 +121,21 @@ object Sub:
   ): Sub[F, A] =
     make[F, A, A, R](id)(acquire)(release)(Option.apply)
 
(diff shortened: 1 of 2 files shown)


gvolpe:
Awesome Dave! I need to run now, but tomorrow I'll have time to properly test it on the trading application and get back to you :)

gvolpe (dismissed):
Works like a charm! 🚀

PR #164, opened 2023-01-15
Sub: add Eq & Functor instances + law checks

gvolpe:
The `Monoid` associative laws are not passing. It could be that the `Eq` instance I wrote is not entirely correct (I don't know if the `id` of `Observe` is the right way to compare two subs?) , or that there is an issue with the `Monoid[Sub]` implementation. 

The following instance yields the same outcome:

```scala
given [F[_], Msg: Eq]: Eq[Sub[F, Msg]] = Eq.fromUniversalEquals
```

Perhaps you have a better idea?


gvolpe:
If we don't look at the content of `Combine`, then all laws pass, but it doesn't look right, does it? 😄 

```scala
  given [F[_], Msg: Eq]: Eq[Sub[F, Msg]] with
    def eqv(x: Sub[F, Msg], y: Sub[F, Msg]): Boolean = (x, y) match {
      case (Sub.None, Sub.None)                             => true
      case (Sub.Observe(id1, _, _), Sub.Observe(id2, _, _)) => id1 === id2
      case (Sub.Combine(x1, y1), Sub.Combine(x2, y2))       => true // eqv(x1, x2) && eqv(y1, y2)
      case (Sub.Batch(xs), Sub.Batch(ys)) =>
        xs.size === ys.size && xs.zip(ys).forall((x, y) => eqv(x, y))
      case _ => false
    }
```

armanbilge:
@gvolpe can you share the output of the law failure?

gvolpe:
Complete logs on the previous run: https://github.com/PurpleKingdomGames/tyrian/actions/runs/3923472574/jobs/6707104827

```scala
failing seed for monoid.associative is AwxoYtLVgxbb7SGjeVcntk5yQaHpLoW1ygsjeuxy4cI=
==> X tyrian.SubLawsTests.Monoid[Sub]: monoid.associative 0.02s munit.FailException: /home/runner/work/tyrian/tyrian/tyrian/js/src/test/scala/tyrian/SubLawsTest.scala:26
25:  checkAll("Functor[Sub]", FunctorTests[Sub[IO, *]].functor[Int, Double, String])
26:  checkAll("Monoid[Sub]", MonoidTests[Sub[IO, String]].monoid)
27:

Failing seed: AwxoYtLVgxbb7SGjeVcntk5yQaHpLoW1ygsjeuxy4cI=
You can reproduce this failure by adding the following override to your suite:

  override val scalaCheckInitialSeed = "AwxoYtLVgxbb7SGjeVcntk5yQaHpLoW1ygsjeuxy4cI="

Falsified after 0 passed tests.
> Labels of failing property: 
Expected: Combine(Observe(,IO(...),<function1>),Combine(Observe(,IO(...),<function1>),Observe(,IO(...),<function1>)))
Received: Combine(Combine(Observe(,IO(...),<function1>),Observe(,IO(...),<function1>)),Observe(,IO(...),<function1>))
> ARG_0: Observe(,IO(...),<function1>)
> ARG_1: Observe(,IO(...),<function1>)
> ARG_2: Observe(,IO(...),<function1>)
```

tyrian/js/src/test/scala/tyrian/SubLawsTest.scala
```diff
@@ -0,0 +1,28 @@
...
+
+  given [A: Arbitrary]: Arbitrary[Sub[IO, A]] =
+    Arbitrary(Arbitrary.arbitrary[A].map(Sub.emit))
+
+  given Cogen[Sub[IO, Int]] = Cogen {
+    case Sub.None             => 0L
+    case Sub.Observe(_, _, _) => 1L
+    case Sub.Combine(_, _)    => 2L
+    case Sub.Batch(_)         => 3L
+  }
```
armanbilge, line 22: This cogen instance is suspect. You should derive it also from the contents, not just the type.

gvolpe: Would something like this make sense? I could never fully understand `Cogen` 😄 

```scala
  val cogenValue: Sub[IO, Int] => Long =
    case Sub.None             => 0L
    case Sub.Observe(_, _, _) => 1L
    case Sub.Combine(x, y)    => 2L + cogenValue(x) + cogenValue(y)
    case Sub.Batch(xs)        => xs.foldLeft(3L)((acc, sub) => acc + cogenValue(sub))

  given Cogen[Sub[IO, Int]] = Cogen(cogenValue)
```

davesmith00000:
I expect that if you change the Monoid instance to use Batch instead of Combine, you'll find the law check passes, because it'll form a monoid on the basis that list is a monoid.

Combine isn't associative because these structures clearly don't equate (I assume, this isn't something I look into much):

```
Combine(Combine(a, b), c) =!= Combine(a, Combine(b, c))
```

gvolpe:
@davesmith00000 do you mean doing this?

```scala
  final def merge[F[_], Msg, LubMsg >: Msg](a: Sub[F, Msg], b: Sub[F, LubMsg]): Sub[F, LubMsg] =
    (a, b) match {
      case (Sub.None, Sub.None) => Sub.None
      case (Sub.None, s2)       => s2
      case (s1, Sub.None)       => s1
      case (s1, s2)             => Sub.Batch(s1, s2)
    }
```

This isn't passing associativity either :/

```scala
failing seed for monoid.associative is _q2a0WOV17VQG2T-uuiwqK5-8BxvNQvW5ZXKG5VllcO=
==> X tyrian.SubLawsTests.Monoid[Sub]: monoid.associative 0.02s munit.FailException: /home/gvolpe/workspace/tyrian-fork/tyrian/js/src/test/scala/tyrian/SubLawsTest.scala:27
26:  checkAll("Functor[Sub]", FunctorTests[Sub[IO, *]].functor[Int, Double, String])
27:  checkAll("Monoid[Sub]", MonoidTests[Sub[IO, String]].monoid)
28:
Failing seed: _q2a0WOV17VQG2T-uuiwqK5-8BxvNQvW5ZXKG5VllcO=
You can reproduce this failure by adding the following override to your suite:

  override val scalaCheckInitialSeed = "_q2a0WOV17VQG2T-uuiwqK5-8BxvNQvW5ZXKG5VllcO="

Falsified after 0 passed tests.
> Labels of failing property:
Expected: Batch(List(Observe(,IO(...),<function1>), Batch(List(Observe(,IO(...),<function1>), Observe(,IO(...),<function1>)))))
Received: Batch(List(Batch(List(Observe(,IO(...),<function1>), Observe(,IO(...),<function1>))), Observe(,IO(...),<function1>)))
> ARG_0: Observe(,IO(...),<function1>)
> ARG_1: Observe(,IO(...),<function1>)
> ARG_2: Observe(,IO(...),<function1>)
```

tyrian/js/src/test/scala/tyrian/SubLawsTest.scala
armanbilge: The `Cogen` implementation should be matching your notion of equality. Usually I implement it by delegating to the cogen of other simpler structures.

Essentially can want to unpack a `Sub` like this:
```scala
Either[
  Unit, // Sub.None
  Either[
    Int, // Sub.Observe id
    Either[
      (Sub, Sub), // Sub.Combine
      List[Sub] // Sub.Batch
    ]
  ]
]
```
And then get the cogen of that.

davesmith00000:
No that won't do it, you'll have to be sneakier. :smile:

Essentially all results of a merge will have to produce a `Sub.Batch`, or you'll forever be comparing trees of different shapes that will never be trivially equal.

The alternative is to work on the `Eq` instance so that is essentially disregards the shape of the tree and only concerns itself with the ultimate order of the concrete subs. Essentially `flatten` then check equals.

armanbilge:
> The alternative is to work on the `Eq` instance so that it essentially disregards the shape of the tree and only concerns itself with the ultimate order of the concrete subs. Essentially `flatten` then check equals.

👍 this sounds right.

tyrian/js/src/test/scala/tyrian/SubLawsTest.scala
gvolpe: My brain is not ready for this challenge today 😆 I'll leave it here for now:

```scala
  type SubCogen[A] = Either[
    Unit, // Sub.None
    Either[
      String, // Sub.Observe id
      Either[
        (Sub[IO, A], Sub[IO, A]), // Sub.Combine
        List[Sub[IO, A]]          // Sub.Batch
      ]
    ]
  ]

  val cogenValue: Sub[IO, Int] => SubCogen[Int] =
    case Sub.None              => Left(())
    case Sub.Observe(id, _, _) => Right(Left(id))
    case Sub.Combine(x, y)     => Right(Right(Left(x, y)))
    case Sub.Batch(xs)         => Right(Right(Right(xs)))

  given Cogen[Sub[IO, Int]] = Cogen.cogenEither.contramap(cogenValue)
```

gvolpe:
> The alternative is to work on the Eq instance so that it essentially disregards the shape of the tree and only concerns itself with the ultimate order of the concrete subs. Essentially flatten then check equals.

Sounds good. Now how would I "flatten" a sub? 🤔 Need to implement `FlatMap` now? 😄

davesmith00000:
`Sub`s and `Cmd`s are just trees of values. So all you have to do (other that implement flatMap) is recursively unwrap the tree into a flat `List`. E.g.:

```scala
Combine(Batch(a, b, c), Combine(d, Combine(e, None)))
```
becomes
```scala
List(a, b, c, d, e)
```

davesmith00000:
@gvolpe ...in fact:

https://github.com/PurpleKingdomGames/tyrian/blob/d5dab3faa80a132b87b7dcaac36c6c61d34ca791/tyrian/js/src/main/scala/tyrian/runtime/SubHelper.scala#L12

gvolpe:
Awesome! That helps a lot, got it working now 🥳

davesmith00000:
@gvolpe I'll be doing some work on Tyrian a bit later (trying to get a release together). If you're happy that this is good to go then I'll have a go at merging it?

gvolpe:
@davesmith00000 yes, go for it. I plan to get back to getting that `Cogen` instance right at some point, but I am afraid I won't have time until next weekend at least.

davesmith00000 (approved):
LGTM! :+1:

PR #168, opened 2023-01-22
Cmd: add Eq & Functor instances + law checks

gvolpe:
Follow up to the previous `Sub` work. I initially had this implementation, but then I realized an instance based on the helpers was much easier (and perhaps more correct) to define and understand. 

```scala
  given [A: Cogen]: Cogen[Sub[IO, A]] =
    def f: (Seed, Sub[IO, A]) => Seed =
      case (seed, Sub.None) =>
        Cogen.perturb(seed, ())
      case (seed, Sub.Observe(id, _, _)) =>
        Cogen.perturb(seed, id)
      case (seed, Sub.Combine(x, y)) =>
        (x, y).foldLeft(seed)((s1, sub) => f(s1, sub))
      case (seed, Sub.Batch(xs)) =>
        xs.foldLeft(seed)((s1, sub) => f(s1, sub))

    Cogen[Sub[IO, A]](f)
```

So I went for it 🥇 

```scala
 given [A: Cogen]: Cogen[Sub[IO, A]] =
    Cogen[List[String]].contramap(SubHelper.flatten(_).map(_.id))
```

### DIFF
--- a/tyrian/js/src/main/scala/tyrian/Cmd.scala
+++ b/tyrian/js/src/main/scala/tyrian/Cmd.scala
@@ -1,8 +1,13 @@
 package tyrian
 
+import cats.Applicative
+import cats.Functor
 import cats.effect.kernel.Sync
 import cats.effect.kernel.Temporal
+import cats.kernel.Eq
 import cats.kernel.Monoid
+import cats.syntax.eq.*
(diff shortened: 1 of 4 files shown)


davesmith00000 (approved):
Excellent, thank you!

PR #169, opened 2023-01-27
Express the render loop in the effect

armanbilge:
1. The next animation frame won't be requested until the current rendering completes
2. As a result, the entire process can run on a single fiber, instead of starting a new one for each animation frame

### DIFF
--- a/tyrian/js/src/main/scala/tyrian/runtime/TyrianRuntime.scala
+++ b/tyrian/js/src/main/scala/tyrian/runtime/TyrianRuntime.scala
@@ -107,7 +107,11 @@ final class TyrianRuntime[F[_]: Async, Model, Msg](
 
   // Render
 
-  private def renderLoop(): Unit =
+  private def renderLoop: F[Unit] =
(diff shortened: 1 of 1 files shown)


davesmith00000:
Good Morning @armanbilge, :smile: 

Looks like a harmless refactor with nice implications, however it doesn't seem to work for me, e.g. in the sandbox, clicking the links at the top changes the anchor but the page content stays the same.

It appears to work initially in that the vdom is initialised so everything renders, but `Msg`s aren't getting processed. I saw this behaviour a lot when I was switching to CE the first time around.

I haven't got the time just now to work out what going on unfortunately as I'm heading off to work, but I'll try and have a look later.

armanbilge:
Heh, that was rather arrogant of me to assume it would just work 😅 it _is_ broken 😕 sorry to waste your time.

But I don't understand how such an innocent refactor can completely break things. I checked and the render-loop is running, so it feels like a weird issue of non-local interactions 😕

davesmith00000:
Don't worry about it, no problem at all. If anything, I think I tried something similar once upon a time and it's nice to know that it wasn't just me that tripped over it. 🙂

PR #170, opened 2023-01-29
docs: networking

gvolpe:
Adding documentation for `Http` and `WebSocket`, and a bit about the integration with `http4s-dom`. It can definitely be further improved, but I did what I could in the spare time I had today and it seems better than nothing 😃 

~~One thing to notice is that the `scaladoc`s are currently failing to compile where `Dispatcher.sequential` is used, but I didn't look into it~~ FIXED. The `mdoc` documentation is all compiling, though 👍🏽

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -215,9 +215,14 @@ lazy val jsdocs =
       organization := "io.indigoengine",
       libraryDependencies ++= Seq(
         "org.scala-js"    %%% "scalajs-dom"          % Dependancies.scalajsDomVersion,
+        "io.circe"        %%% "circe-core"           % Dependancies.circe,
+        "io.circe"        %%% "circe-generic"        % Dependancies.circe,
+        "io.circe"        %%% "circe-parser"         % Dependancies.circe,
         "io.indigoengine" %%% "indigo"               % indigoDocsVersion,
         "io.indigoengine" %%% "tyrian-io"            % tyrianDocsVersion,
-        "io.indigoengine" %%% "tyrian-indigo-bridge" % tyrianDocsVersion
+        "io.indigoengine" %%% "tyrian-indigo-bridge" % tyrianDocsVersion,
+        "org.http4s"      %%% "http4s-dom"           % "0.2.3",
+        "org.http4s"      %%% "http4s-circe"         % "0.23.15"
       )
     )
     .enablePlugins(ScalaJSPlugin)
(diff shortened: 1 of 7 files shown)


armanbilge:
> One thing to notice is that the `scaladoc`s are currently failing to compile where `Dispatcher.sequential` is used

The `cats-effect-std` dependency should be bumped to 3.4.0+ to fix this.

build.sbt
```diff
@@ -215,9 +215,14 @@ lazy val jsdocs =
...
       libraryDependencies ++= Seq(
         "org.scala-js"    %%% "scalajs-dom"          % Dependancies.scalajsDomVersion,
+        "io.circe"        %%% "circe-core"           % Dependancies.circe,
+        "io.circe"        %%% "circe-generic"        % Dependancies.circe,
+        "io.circe"        %%% "circe-parser"         % Dependancies.circe,
         "io.indigoengine" %%% "indigo"               % indigoDocsVersion,
         "io.indigoengine" %%% "tyrian-io"            % tyrianDocsVersion,
-        "io.indigoengine" %%% "tyrian-indigo-bridge" % tyrianDocsVersion
+        "io.indigoengine" %%% "tyrian-indigo-bridge" % tyrianDocsVersion,
+        "org.http4s"      %%% "http4s-dom"           % "0.2.3",
```
armanbilge, line 224:
```suggestion
        "org.http4s"      %%% "http4s-dom"           % "0.2.6",
```

docs/concepts/networking.md
```diff
@@ -4,4 +4,286 @@ title: "Networking"
...
+    case Msg1.MorePlease     => (model, HttpHelper.getRandomGif(model.topic))
+    case Msg1.NewGif(newUrl) => (model.copy(gifUrl = newUrl), Cmd.None)
+    case Msg1.GifError(_)    => (model, Cmd.None)
+```
+
+You can find the full code on the examples directory linked at the top.
+
+### Http4s-dom integration
+
+To use `http4s-dom` instead, we only need to replace the `HttpHelper` with the following implementation.
```
armanbilge, line 106: Thanks for adding this! :)

gvolpe:
Thanks for reviewing @armanbilge :)

gvolpe:
It's a shame there's no `scala mdoc:js:reset` -_- It forces you to use different names on the same file (e.g. `Msg1`, `Msg`, etc).

davesmith00000 (approved):
Great work! Thank you! :pray:

PR #171, opened 2023-02-01
Take 2: refactor the entire runtime instead ;)

armanbilge:
Well, I was hoping to work up to this via a series of small changes, but https://github.com/PurpleKingdomGames/tyrian/pull/169 put a damper on that plan.

Believe it or not, the first commit is precisely my first attempt, guided only by the existing implementation and compiler. This time I did have the good sense to try running the sandbox 😜 and remarkably it seems to be less broken than whatever I did in https://github.com/PurpleKingdomGames/tyrian/pull/169 😂

I'm going to fiddle with it some more since some things seem off, but in the long-term I will need help recognizing bugs since I don't have enough intuition about the "Tyrian feel" and expected semantics.

This is a non-trivial refactor, so I expect plenty of questions and slow-and-steady review cycles :)


armanbilge:
> since some things seem off

Well, I can't really tell to be honest 🤔 

I upgraded to a pre-release of Cats Effect 3.5 which is shipping a new runtime for JS. This seems to have helped, and as far as I can tell everything in the sandbox is working. I think 😅

armanbilge:
Yeah, there is definitely something off. It's utterly bizarre actually: it's like the `Channel` completely loses the first handful of messages submitted to it. Maybe this is what's been plaguing us all along.

armanbilge:
Ok, now I think everything is actually working as expected 😅

project/Dependancies.scala
```diff
@@ -1,6 +1,6 @@
 object Dependancies {
 
-  val catsEffect        = "3.4.4"
+  val catsEffect        = "3.5-51ca39e"
```
armanbilge, line 3:
```suggestion
  val catsEffect        = "3.5.0-RC1"
```

davesmith00000:
Are we far off a CE3 3.5.0 release do you know @armanbilge?

armanbilge:
Probably at least a couple weeks to give the RC a chance to percolate, possibly longer.

tyrian/js/src/main/scala/tyrian/runtime/TyrianRuntime.scala
```diff
@@ -1,134 +1,97 @@
...
+
+  def apply[F[_], Model, Msg](
+      node: Element,
+      initModel: Model,
+      initCmd: Cmd[F, Msg],
+      update: Model => Msg => (Model, Cmd[F, Msg]),
+      view: Model => Html[Msg],
+      subscriptions: Model => Sub[F, Msg]
+  )(using F: Async[F]): F[Nothing] = Dispatcher.sequential[F].use { dispatcher =>
+    (F.ref(ModelHolder(initModel, true)), AtomicCell[F].of(List.empty[(String, F[Unit])]), Channel.unbounded[F, Msg])
```
davesmith00000, line 31: For my education, why not use an `AtomicCell` for the model too?

tyrian/js/src/main/scala/tyrian/runtime/TyrianRuntime.scala
```diff
@@ -1,134 +1,97 @@
...
+      .flatMapN { (model, currentSubs, msgs) =>
```
davesmith00000, line 32: Nitpicking, but `msgs` sounds like the result rather than the conduit and forces a bit of mental gymnastics.

Later you're doing `msgs.send(_)` which reads like "Ok yes, send the messages... no wait... the message was in the foreach, this is sending those _into_ msgs...right?"

tyrian/js/src/main/scala/tyrian/runtime/TyrianRuntime.scala
```diff
@@ -1,134 +1,97 @@
...
+
+        def runCmd(cmd: Cmd[F, Msg]): Stream[F, Nothing] =
+          Stream
+            .emits(CmdHelper.cmdToTaskList(cmd))
+            .parEvalMapUnorderedUnbounded(_.handleError(_ => None))
+            .unNone
```
davesmith00000, line 38: TIL - no more collecting Some's for me! :smile:

tyrian/js/src/main/scala/tyrian/runtime/TyrianRuntime.scala
armanbilge: Good question. A `Ref` can only be updated/modified with pure functions, which is sufficient for the model.

The logic for updating `currentSubs` is more complicated since it involves effects. An `AtomicCell` can be modified with an effectual function, but to implement this safely it adds a small amount of overhead.

tyrian/js/src/main/scala/tyrian/runtime/TyrianRuntime.scala
armanbilge: Not nitpicking at all, you are right :) `msgChannel` maybe?

davesmith00000:
This very seems to work just fine - thanks @armanbilge! I'm just wandering through the details.

tyrian/js/src/main/scala/tyrian/runtime/TyrianRuntime.scala
```diff
@@ -1,134 +1,97 @@
...
+          model
+            .modify { case ModelHolder(oldModel, _) =>
+              val (newModel, cmd) = update(oldModel)(msg)
+              val sub             = subscriptions(newModel)
+              (ModelHolder(newModel, true), (cmd, sub))
+            }
+            .flatMap { (cmd, sub) =>
+              runSub(sub).as(runCmd(cmd))
+            }
+        }.parJoinUnbounded
```
davesmith00000, line 68: We're ok to do this in parallel because JS is single threaded?

armanbilge: Hm, actually I think this is necessary for correct semantics. Because otherwise a side-effect like an `IO.sleep(...)` or some network request could block processing of other cmds/subs.

By running in parallel, it's enabling other work to be done while some tasks are asynchronous suspended.

tyrian/js/src/main/scala/tyrian/runtime/TyrianRuntime.scala
davesmith00000: Ah, excellent. That probably would have been clear if I tried to use them myself but wasn't obvious from just reviewing the code. Thanks!

tyrian/js/src/main/scala/tyrian/runtime/TyrianRuntime.scala
davesmith00000: Yes you're right, I see what's going on now - thanks. Just thinking about this out loud:

When _I_ think about the update loop I tend to think of the loop - quite wrongly - that the messages and effects are strictly ordered, and they are emitted in order, but of course that's wrong and they may take variable lengths of time to arrive if - as you say - they contained a callback or a sleep or something.

In practice the fact that messages and their results don't resolve in their original order is a) not a real problem* and b) impractical because it would mean blocking on everything.

(*The way you end up programming things in Tyrian / Elm means that either you're waiting for your own messages to trigger an onward journey, thus order is enforced by design, or order doesn't matter and you just want everything to happen as quickly as possible.)

davesmith00000:
LGTM - and it's still pretty straightforward I think. Most of my difficulty following the changes is just having to stop and read up on CE stuff. :smile: 

I may hang on for the next CE release before I merge (perhaps?), but looks great! Thanks again!

tyrian/js/src/main/scala/tyrian/runtime/TyrianRuntime.scala
armanbilge: The * is a nice observation :)

davesmith00000:
We had a chat on Discord and I'm trying to recall the outcome, I think we left it here:

image
https://discord.com/channels/716435281208672356/851130636458655754/1086956152060911708

Which means I believe this is on hold until the CE 3.5.0 release, agreed?

davesmith00000:
Thanks for updating this @armanbilge. I'm in the middle of some work on Tyrian and I'm behind on PRs, will merge soon!

davesmith00000:
As discussed on Discord, I've done a little checking over the performance. This is more or less a copy+paste of the discord message:

Memory under the refactor is less stable, but only uses about half that of the current version.
profiling_mem_main_2023-05-14_08-31
profiling_mem_runtime_refactor_2023-05-14_08-31

All things are a little quicker and lighter in the new version. One exception is the actual task length which is weirdly twice as long.

profiling_cpu_main_2023-05-14_08-31
profiling_cpu_runtime_refactor_2023-05-14_08-31

A _possible_ reason for that is that `requestAnimationFrame` seems to get called twice consistently.

image

This doesn't seem to happen on the version on `main` branch, and (if it isn't a red herring), presumably means your version could be even faster.

davesmith00000:
It's a red herring! The sandbox app that I was using for comparison also invokes a `requestAnimationFrame` and that is where the double up comes from. If you disable that then it's super snappy. :muscle: 

I'll raise a separate issue though because it would be nice to (unrelated to this PR) be able to hook into the main render loop rather than requesting animation frames repeatedly.

So I think.. it's good?

tyrian/js/src/main/scala/tyrian/TyrianAppF.scala
```diff
@@ -111,14 +111,13 @@ trait TyrianAppF[F[_]: Async, Msg, Model]:
 
   def ready(node: Element, flags: Map[String, String]): Unit =
     run(
-      Tyrian.start(
+      Tyrian.start[F, Model, Msg](
         node,
         router,
         _init(flags),
```
davesmith00000, line 117: I messed up the process somehow here. I thought I was going to suggest a change but I've committed it to your branch via the GH UI - didn't even know that was a thing!

Anyway, you have reverted `_init` to `init` and that was preventing the app from routing when it first opens. So if you were on 'page 2' visited an external link, hit the back button, you'd find yourself on the homepage instead of 'page 2' even though the address bar says `/page-2`.

davesmith00000: The good news is that with this change, the app loads significantly more quickly and the initial page route happens before the first render, removing an initial flicker than I was going to investigate as a separate issue. :muscle:

davesmith00000:
I was about to merge this, but I've found a bug, here's the console exception:

image

And the page doesn't render correctly as you might expect.

Note: This works correctly on `main` branch.

As per my other comment, I reinstated the use of `_init` in the `TyrianAppF` file, and this is when the bug appears in the ZIO based indigo-sandbox.

If you use `init` (not `_init`) then there is no bug but initial routing doesn't work. So `_init` is back. The indigo sandbox uses the Routing helper `externalOnly` which, long story short, produces a `Cmd.None` in the model update for internal links. If you replace that `Cmd.None` with `Nav.pushUrl(the href)` then the exception doesn't throw and everything works.

Ah ha! Must by some combination of routing and resulting in a `Cmd.None` then! Doesn't look like it though, because that's what the `IO` sandbox does for internal links and it works just fine.

I'm out of time to dig into this now - but my guess is that something isn't completing or being run or something - maybe to do with the ZIO runtime stuff? Maybe adding the `Nav.pushUrl` is just forcing the cmd's to evaluate? Just guessing.

davesmith00000:
I've had a look into this and I've requested help here: https://github.com/zio/zio/issues/8125

davesmith00000:
Got it @armanbilge - https://github.com/PurpleKingdomGames/tyrian/pull/171/commits/d32d7744dea25a944d2acac5e8b8acf0145446e5

I was slightly mistaken. When I tried switching to `fork` I saw the page load failing and assumed (:facepalm:) that it was the same issue. In fact moving to `fork` works, but we need to be a bit smarter about how the Indigo parts of the test app are launched.

davesmith00000 (approved):

PR #175, opened 2023-03-13
Use cats.effect.std.Queue as the WebSocketEvent queue.

uncle-betty:
Hello there.

I'd like to propose a change of data structure for the `WebSocketEvent` queue. The current implementation is based on a `Stream` that busy-polls a `scala.collection.mutable.Queue`. The resulting tight loop uses one full CPU core on my laptop, which makes me worry a little about the battery life of mobile devices that access my Tyrian app.

I'd like to explore changing the queue to a blocking `cats.effect.std.Queue`. I have given this a try (and things seem to work), but I'm not entirely sure that I know what I'm doing. I'm still new to Cats and friends. So, I'd like some adult supervision, please, if at all possible.

The changes seem to do the job for me. But they look a little weird, don't they? To elaborate: As before, we have imperative code that `newConnection` wraps in an `Async[F].delay`. So, that's fine. However, the imperative code now contains a function, `enqueue`, that spins up a `Dispatcher`, so that it can do the blocking via `unsafeRunAndForget()`. But then this part of `enqueue` is, in turn, wrapped in an `Async[F].delay`, because the `Resource`'s `use` want's an `F`.

That seems like a lot of back and forth between imperative code and effectful code, doesn't it? But I do need the `unsafeRunAndForget()`, don't I?

Anyway. Just a proposal. Feel free to close the pull request, if this isn't a priority or if my idea is completely ill-informed.

Thank you for making Tyrian. It's been an absolute pleasure to work with.

### DIFF
--- a/tyrian/js/src/main/scala/tyrian/websocket/WebSocket.scala
+++ b/tyrian/js/src/main/scala/tyrian/websocket/WebSocket.scala
@@ -2,6 +2,8 @@ package tyrian.websocket
 
 import cats.effect.kernel.Async
 import cats.effect.kernel.Sync
+import cats.effect.std.Dispatcher
+import cats.effect.std.Queue
 import cats.syntax.functor.*
 import fs2.Stream
 import org.scalajs.dom
@@ -10,7 +12,6 @@ import tyrian.Sub
 import tyrian.websocket.WebSocketEvent
 import util.Functions
 
-import scala.collection.mutable.Queue
 import scala.concurrent.duration.*
 
 /** Helper WebSocket instance to store in your model that acts as a controller */
@@ -105,28 +106,36 @@ object WebSocket:
     Async[F].delay {
       val socket    = new dom.WebSocket(address)
       val keepAlive = new KeepAlive(socket, settings)
-      val q         = Queue.empty[WebSocketEvent]
+      val q         = Queue.unbounded[F, WebSocketEvent]
+
+      val enqueue: WebSocketEvent => Unit = wse =>
+        Dispatcher.sequential[F].use { disp =>
+          Async[F].delay {
+            val action = Async[F].flatMap(q)(_.offer(wse))
+            disp.unsafeRunAndForget(action)
+          }
+        }
 
       val msgListener = Functions.fun { e =>
         val event = WebSocketEvent.Receive(e.asInstanceOf[dom.MessageEvent].data.toString)
-        q.enqueue(event)
+        enqueue(event)
       }
 
       val errListener = Functions.fun { e =>
         val msg =
           try e.asInstanceOf[dom.ErrorEvent].message
           catch { case _: Throwable => "Unknown" }
-        q.enqueue(WebSocketEvent.Error(msg))
+        enqueue(WebSocketEvent.Error(msg))
       }
 
       val closeListener = Functions.fun { e =>
         val ev = e.asInstanceOf[dom.CloseEvent]
-        q.enqueue(WebSocketEvent.Close(ev.code, ev.reason))
+        enqueue(WebSocketEvent.Close(ev.code, ev.reason))
       }
 
       val openListener = Functions.fun { _ =>
         onOpenSendMessage.foreach(msg => socket.send(msg))
-        q.enqueue(WebSocketEvent.Open)
+        enqueue(WebSocketEvent.Open)
       }
 
       socket.addEventListener("message", msgListener)
@@ -138,11 +147,8 @@ object WebSocket:
         Sub.Batch(
           Sub.make(s"[tyrian-ws-${address}]")(
             Stream.repeatEval {
-              Sync[F].delay {
-                try Some(q.dequeue())
-                catch _ => None
-              }
-            }.unNone
+              Async[F].flatMap(q)(_.take)
+            }
           )(
             Async[F].delay {
               socket.removeEventListener("message", msgListener)


uncle-betty:
Oops. I'm afraid that I have to retract this pull request. The changes now do not seem to work. I don't know what happened or what I tested before.

I'll come back once I've figured this out.

tyrian/js/src/main/scala/tyrian/websocket/WebSocket.scala
```diff
@@ -105,28 +106,36 @@ object WebSocket:
...
-      val q         = Queue.empty[WebSocketEvent]
+      val q         = Queue.unbounded[F, WebSocketEvent]
+
+      val enqueue: WebSocketEvent => Unit = wse =>
+        Dispatcher.sequential[F].use { disp =>
+          Async[F].delay {
+            val action = Async[F].flatMap(q)(_.offer(wse))
+            disp.unsafeRunAndForget(action)
+          }
+        }
```
armanbilge, line 117: I think I know why your PR didn't work :) there's a couple issues with this code:
1. it's opening/closing the dispatcher (the `use`) before the submitted action has a chance to run
2. you are never actually running this effect!

The motivation behind this PR is excellent btw, we just need to adjust the approach a bit. Basically, you want to create a `Dispatcher` once, and re-use it for all events. Here's an example.

https://github.com/http4s/http4s-dom/blob/8734cdc909063a157bca1f95b625032bb5e63cab/dom/src/main/scala/org/http4s/dom/WebSocketClient.scala

davesmith00000:
I think my queue knowledge is failing me a bit here (or maybe I just need my morning coffee :smile:) but I though the original (and good) observation was that the busy polling was bad for resource utilisation? How does changing the queue type help with that situation?

The change you're looking at was put in recently to fix a bug, and I worried about it a bit as a solution so I'm pleased to explore alternatives.

armanbilge:
> How does changing the queue type help with that situation?

I haven't looked at the old/current implementation, but the Cats Effect queue is _not_ based on busy polling. It uses asynchronous callbacks.

davesmith00000:
> > How does changing the queue type help with that situation?
> 
> I haven't looked at the old/current implementation, but the Cats Effect queue is _not_ based on busy polling. It uses asynchronous callbacks.

I've been mulling this over since early this morning, and I've come to the conclusion that I have no idea what either of you are talking about, and that the best thing I can do is ask an ignorant question in the hope of later enlightenment. 😁

Why would any queue - which is a data structure - have a busy polling implementation attached to it?

Do you mean that fs2 is using busy polling when using a standard mutable queue as a source in order to checked for queued items? And therefore uses a different strategy with a CE Queue, perhaps?

armanbilge:
> Why would any queue - which is a data structure - have a busy polling implementation attached to it?

Well, cause somebody wrote this code 😛 

https://github.com/PurpleKingdomGames/tyrian/blob/bf51869ad6bf361ebe7322e67a7ed7df675ed848/tyrian/js/src/main/scala/tyrian/websocket/WebSocket.scala#L140-L145

You can see it there, that's running a hard loop that repeatedly tries to dequeue something, and if it doesn't get anything, it simply loops again immediately. Hence, busy-polling.

The Cats Effect `Queue` does not work like this. Instead, it registers an asynchronous callback, so a listener can be awakened when new data is available in the queue. Until then, the listener is "sleeping"/idle.

uncle-betty:
Oooooooh... Right! Yes! Thank you, @armanbilge. It helps, when you run the effects that you create.

For some reason, I thought that `use` returns a `Unit`, i.e., that it runs the effect. But it doesn't. It returns an `F[Unit]`.

Note that the Scala type system didn't catch that: My `enqueue` function is declared as returning `Unit`, not `F[Unit]`. I guess that Scala just allows arbitrarily typed last expressions in functions that return `Unit`. Which makes sense.

As to why I thought my solution worked: I accidentally had my client-side code make HTTP requests to the server, in addition to the websocket. So, the data that I saw on the client side had actually arrived via the HTTP requests, not via the websocket. So, the websocket never worked.

armanbilge:
> Note that the Scala type system didn't catch that

It will, if you ask it to :)

`-Wvalue-discard` / `Ywarn-value-discard`
> Warn when nullary methods return Unit.

https://docs.scala-lang.org/overviews/compiler-options/index.html

uncle-betty:
Regarding @davesmith00000's question: `Stream.repeatEval` creates a fiber that repeatedly runs the given effect. It does so in a tight loop. Suppose that the queue is empty. Now:

* `repeatEval` runs the effect. Which does the `q.dequeue()`. Which causes an exception. Which produces a `None`. Which gets discarded by the `unNone`.
* Then `repeatEval` runs the effect again, and the same thing happens.
* And so on, and so on.

I'm using a single websocket and it saturates a single CPU core. Using 8 websockets would saturate all 8 CPUs of my laptop.

In contrast, when running `take` on an empty `cats.effect.std.Queue`, then the fiber that runs the `take` is put to sleep until the queue becomes non-empty.

uncle-betty:
Ooooooh... Thank you, @armanbilge, for the compiler option. Great! That'll keep me safe in the future.

uncle-betty:
Regarding a fix: I don't think that we can easily thread a `Dispatcher` all the way to the websocket creation, can we?

Given that it's the user of Tyrian who calls `WebSocket.connect`, wouldn't they have to pass the `Dispatcher` to `WebSocket.connect` and we'd then thread it down from there to `newConnection`, where it's needed?

uncle-betty:
Oh, and to add to @davesmith00000's question: Maybe another way of looking at this is that we're dealing not just with a queue as a data structure, but an implementation of this data structure.

And and implementation of a queue can either block when a fiber / thread / process tries to grab an element from an empty queue (until an element to be returned is available) or it can return an error / exception / None to the fiber / thread / process indicating that it's empty.

`cats.effect.std.Queue` does the former. `scala.collection.mutable.Queue` does the latter.

uncle-betty:
Ooooh, OK, I think that I see a little clearer now as to what you, Arman, proposed. I took a closer look at the code you linked to.

So, what you're proposing is to turn the websocket into a `Resource`, right? We'd nest that into a `Dispatcher` resource and thus have a `Dispatcher` at our disposal for the lifetime of the websocket. Clever!

I'll think about this a little more. Maybe it's within my means to give this a try.

davesmith00000:
> You can see it there, that's running a hard loop that repeatedly tries to dequeue something

Yep, makes total sense, thanks. That's more or less what I was asking. :+1:

davesmith00000:
> `-Wvalue-discard` / `Ywarn-value-discard`

I'm surprised they aren't in sbt-tpolecat? Although maybe I'm a few versions behind or something...

davesmith00000:
I think the catch here is going to be that Tyrian's API's don't currently expose any means to make use of a `Resource`, or to supply a `Dispatcher`. I'm not saying we can't do that, but it may have API implications.

@uncle-betty, if you can make it work by any means - however ugly! :yum: - we can think about how to make it nice afterwards.

armanbilge:
> I think the catch here is going to be that Tyrian's API's don't currently expose any means to make use of a `Resource`

I think they do, at least indirectly :) so I remember that I had a PR that lets you convert an `fs2.Stream` to a `Sub`.
- https://github.com/PurpleKingdomGames/tyrian/pull/140

A `Dispatcher` or actually any `Resource` can be lifted into a `Stream` with `Stream.resource(...)`. So that seems like that could work.

davesmith00000:
Hmmmm. You're right, but I still think you might have issues. We'll see, maybe it's fine. (I hope it is!)

That queue is specifically there because of a race condition where the websocket is ready and firing before the Sub has been initialised.

Another solution to this problem is that once the initial messages have been processed and you're into normal operation, you can actually skip / disable / by-pass the queue entirely.

armanbilge:
> That queue is specifically there because of a race condition where the websocket is ready and firing before the Sub has been initialised.

Hmm, why can't the `WebSocket` be created as part of the Sub, something to do with being able to send messages on it?

davesmith00000:
Here's the issue where we discussed / diagnosed the problem.

https://github.com/PurpleKingdomGames/tyrian/issues/149

armanbilge:
> you need a reference to the socket available in your model

If the websocket is going in the model, then the dispatcher and queue can go into the model too 😁 

The big picture here is that a websocket is a resource (i.e. it is something to be closed) and a dispatcher is also a resource. So wherever one goes, the other should be able to go as well.

davesmith00000:
Slippery slope, because the implication is that I imagine you'll want an `F[Model]` instead of just a `Model`, or something, and that isn't appealing to me.

armanbilge:
Huh? not sure I follow :)

Do you mean that creating a `Dispatcher` is an effect? Well, so is creating a `WebSocket` 😁 in fact, it's _already_ wrapped in a `delay` block.

davesmith00000:
Yeah I started having second thoughts about that comment after I sent it. ☺️🤷

We could try it and see how it went.

The thing I specifically want to avoid is the need to weave effect monads through everything. At the moment, they're there, but they are nicely kept at arm's length at the edge of the application where the side effecting work happens (in the subs and cmds).

If we ever got to the point where the return type of the update function was `F[(Model[F], CMD[F, Msg]]` or some such, then we've lost the game and we should pack up and use something else. 😄

armanbilge:
> The thing I specifically want to avoid is the need to weave effect monads through everything

Sure, I mean there is an easy fix to this, which is basically the better version of what you've done here :) instead of using an effectful queue, you can use an ad-hoc mutable queue. The problem is the one used here doesn't support async notifications, so the implemented was based on busy-polling. But it wouldn't be too hard to implement an (unsafe) mutable, async queue, that doesn't have any of those effect monads.

In fact, you could copy it right out of Cats Effect :) since that's literally the thing _under_ those effects monads.

PR #176, opened 2023-03-16
Use a blocking queue for WebSocketEvents.

uncle-betty:
Hello there.

I would like to explore a new draft proposal, which actually solves the issue for me. It's not a particularly elegant abstraction and it also has a few known issues, but on the other hand it's a pretty minimal set of changes that are invisible from the user's perspective as far as I can tell. Maybe it can be turned into or inspire something useful.

* I added a new command, `BuildRun`. It wraps a function, `build`, that basically produces a `Run` from a `Dispatcher`. That's what `WebSocket.connect` now returns instead of a `Run`.
* `cmdToTaskList` now takes a `Dispatcher` so that it can deal with `BuildRun`s: apply `build` to the `Dispatcher`, then treat the result just like a `Run`.
* Inside the `BuildRun`, the `Dispatcher` is threaded down to `newConnection`, which uses it to push elements on the queue.

Unresolved issues (see the `XXX` marks in the code):

* How would one now define equality on commands? For now, I've commented it out.
* There are two type casts in `cmdToTaskList`, in the code that processes the `BuildRun`. I don't understand why I need them. Without them, Scala doesn't see that, for example, the `F` in the type of `build` is identical to the `F` in `cmdToTaskList`'s type signature.
* Is there a way to bring `flatMap` into scope so that we can use a for-comprehension in `newConnection`?

Oh, and some of the lines have gotten really long. Is there a style guide for the Tyrian code?

### DIFF
--- a/tyrian/js/src/main/scala/tyrian/Cmd.scala
+++ b/tyrian/js/src/main/scala/tyrian/Cmd.scala
@@ -4,6 +4,7 @@ import cats.Applicative
 import cats.Functor
 import cats.effect.kernel.Sync
 import cats.effect.kernel.Temporal
+import cats.effect.std.Dispatcher
 import cats.kernel.Eq
 import cats.kernel.Monoid
 import cats.syntax.eq.*
@@ -69,6 +70,14 @@ object Cmd:
     def apply[F[_], A, Msg](task: F[A])(toMessage: A => Msg): Run[F, A, Msg] =
       Run(task, toMessage)
 
+  /** Builds and runs a task that needs a Dispatcher */
+  final case class BuildRun[F[_], A, Msg](
+      build: Dispatcher[F] => F[A],
+      toMsg: A => Msg
+  ) extends Cmd[F, Msg]:
+    def map[OtherMsg](f: Msg => OtherMsg): BuildRun[F, A, OtherMsg] =
+      BuildRun(build, toMsg andThen f)
+
   /** Merge two commands into a single one */
   case class Combine[F[_], Msg](cmd1: Cmd[F, Msg], cmd2: Cmd[F, Msg]) extends Cmd[F, Msg]:
     def map[OtherMsg](f: Msg => OtherMsg): Combine[F, OtherMsg] = Combine(cmd1.map(f), cmd2.map(f))
@@ -95,9 +104,12 @@ object Cmd:
     def empty: Cmd[F, Msg]                                   = Cmd.None
     def combine(a: Cmd[F, Msg], b: Cmd[F, Msg]): Cmd[F, Msg] = Cmd.merge(a, b)
 
+  // XXX - How to do equality?
+  /*
   given [F[_]: Applicative, Msg: Eq](using ev: Eq[F[Option[Msg]]]): Eq[Cmd[F, Msg]] with
     def eqv(x: Cmd[F, Msg], y: Cmd[F, Msg]): Boolean =
       CmdHelper.cmdToTaskList(x) === CmdHelper.cmdToTaskList(y)
+  */
 
   given [F[_]]: Functor[Cmd[F, *]] with
     def map[A, B](fa: Cmd[F, A])(f: A => B): Cmd[F, B] = fa.map(f)
--- a/tyrian/js/src/main/scala/tyrian/runtime/CmdHelper.scala
+++ b/tyrian/js/src/main/scala/tyrian/runtime/CmdHelper.scala
@@ -1,15 +1,17 @@
 package tyrian.runtime
 
 import cats.Applicative
+import cats.effect.std.Dispatcher
 import tyrian.Cmd
 
 import scala.annotation.tailrec
 
 object CmdHelper:
 
-  def cmdToTaskList[F[_]: Applicative, Msg](cmd: Cmd[F, Msg]): List[F[Option[Msg]]] =
+  def cmdToTaskList[F[_]: Applicative, Msg](cmd: Cmd[F, Msg],
+    disp: Dispatcher[F]): List[F[Option[Msg]]] =
     @tailrec
-    def rec(remaining: List[Cmd[F, Msg]], acc: List[F[Option[Msg]]]): List[F[Option[Msg]]] =
+    def rec[A](remaining: List[Cmd[F, Msg]], acc: List[F[Option[Msg]]]): List[F[Option[Msg]]] =
       remaining match
         case Nil =>
           acc.reverse
@@ -27,6 +29,12 @@ object CmdHelper:
 
             case Cmd.Run(task, f) =>
               rec(cmds, Applicative[F].map(task)(p => Option(f(p))) :: acc)
+
+            case Cmd.BuildRun(build, f) =>
+              // XXX - How to get rid of these casts?
+              val castBuild = build.asInstanceOf[Dispatcher[F] => F[A]]
+              val castF = f.asInstanceOf[A => Msg]
+              rec(cmds, Applicative[F].map(castBuild(disp))(p => Option(castF(p))) :: acc)
 
             case Cmd.Combine(cmd1, cmd2) =>
               rec(cmd1 :: cmd2 :: cmds, acc)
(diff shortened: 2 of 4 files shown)


davesmith00000:
> Oh, and some of the lines have gotten really long. Is there a style guide for the Tyrian code?

No, and I don't think scalafmt has been enforced in CI yet (I've done this on other projects), but if you can make sure you let scalafmt run on the files you're changed, that would be much appreciated.

tyrian/js/src/main/scala/tyrian/websocket/WebSocket.scala
```diff
@@ -57,104 +58,105 @@ object KeepAliveSettings:
...
-          try e.asInstanceOf[dom.ErrorEvent].message
-          catch { case _: Throwable => "Unknown" }
-        q.enqueue(WebSocketEvent.Error(msg))
-      }
-
-      val closeListener = Functions.fun { e =>
-        val ev = e.asInstanceOf[dom.CloseEvent]
-        q.enqueue(WebSocketEvent.Close(ev.code, ev.reason))
-      }
+    // XXX - How to get flatMap into scope for a for-comprehension?
```
armanbilge, line 108: `import cats.syntax.all._`

tyrian/js/src/main/scala/tyrian/websocket/WebSocket.scala
```diff
@@ -57,104 +58,105 @@ object KeepAliveSettings:
...
+
+        socket.addEventListener("message", msgListener)
+        socket.addEventListener("error", errListener)
+        socket.addEventListener("open", openListener)
+        socket.addEventListener("close", closeListener)
+
+        val subs =
+          Sub.Batch(
+            Sub.make(s"[tyrian-ws-${address}]")(
+              Stream.repeatEval {
```
davesmith00000, line 144: So what we're saying is that although we're still using `repeatEval`, `take` will block fiber when it's empty, effectively preventing a busy loop - is that about right?

tyrian/js/src/main/scala/tyrian/Cmd.scala
```diff
@@ -69,6 +70,14 @@ object Cmd:
     def apply[F[_], A, Msg](task: F[A])(toMessage: A => Msg): Run[F, A, Msg] =
       Run(task, toMessage)
 
+  /** Builds and runs a task that needs a Dispatcher */
+  final case class BuildRun[F[_], A, Msg](
+      build: Dispatcher[F] => F[A],
+      toMsg: A => Msg
+  ) extends Cmd[F, Msg]:
+    def map[OtherMsg](f: Msg => OtherMsg): BuildRun[F, A, OtherMsg] =
+      BuildRun(build, toMsg andThen f)
```
davesmith00000, line 79: I'm not hugely keen on the name `BuildRun` (what does that mean? why does is speak the need for a `Dispatcher`?), but the implementation is interesting, I wouldn't have thought of that.

I need to have a think about this a bit. I feel like its opening a can of worms but I'm not sure how big the can is yet. :smile:

tyrian/js/src/main/scala/tyrian/Cmd.scala
```diff
@@ -95,9 +104,12 @@ object Cmd:
     def empty: Cmd[F, Msg]                                   = Cmd.None
     def combine(a: Cmd[F, Msg], b: Cmd[F, Msg]): Cmd[F, Msg] = Cmd.merge(a, b)
 
+  // XXX - How to do equality?
+  /*
   given [F[_]: Applicative, Msg: Eq](using ev: Eq[F[Option[Msg]]]): Eq[Cmd[F, Msg]] with
     def eqv(x: Cmd[F, Msg], y: Cmd[F, Msg]): Boolean =
       CmdHelper.cmdToTaskList(x) === CmdHelper.cmdToTaskList(y)
+  */
```
davesmith00000, line 112: I take it this doesn't work now, what does the error look like?

davesmith00000 (commented):
Looks interesting! I've left a couple of comments, and I need to have a think. The build is failing around the equality stuff you mentioned.

tyrian/js/src/main/scala/tyrian/websocket/WebSocket.scala
davesmith00000: I'm not sure you need to switch to a for-comp though in all honesty. Looks fine.

tyrian/js/src/main/scala/tyrian/websocket/WebSocket.scala
armanbilge: Actually this can just be `Stream.fromQueueUnterminated(q)` 🙁 

You might also look at `Channel` instead, it is optimized for this specific use-case.
https://github.com/typelevel/fs2/blob/509d74fdcc5bf0735cabcf829d0a5ba21a949ed9/core/shared/src/main/scala/fs2/concurrent/Channel.scala#L29-L31

tyrian/js/src/main/scala/tyrian/runtime/CmdHelper.scala
```diff
@@ -1,15 +1,17 @@
...
+import cats.effect.std.Dispatcher
 import tyrian.Cmd
 
 import scala.annotation.tailrec
 
 object CmdHelper:
 
-  def cmdToTaskList[F[_]: Applicative, Msg](cmd: Cmd[F, Msg]): List[F[Option[Msg]]] =
+  def cmdToTaskList[F[_]: Applicative, Msg](cmd: Cmd[F, Msg],
+    disp: Dispatcher[F]): List[F[Option[Msg]]] =
```
armanbilge, line 12: Our PRs are now conflicting! 😁 
- https://github.com/PurpleKingdomGames/tyrian/pull/171

armanbilge:
Throwing my hat in the ring 🎩 
- https://github.com/PurpleKingdomGames/tyrian/pull/177

tyrian/js/src/main/scala/tyrian/websocket/WebSocket.scala
uncle-betty: Oh, great! Thanks, Arman. That'll be handy in the future. Leaving it as is in this case, as per Dave's comment.

tyrian/js/src/main/scala/tyrian/websocket/WebSocket.scala
uncle-betty: @davesmith00000 - Yes, that's exactly right.

@armanbilge - Oh, cool! Thank you for the pointer. I'll take a look.

tyrian/js/src/main/scala/tyrian/Cmd.scala
uncle-betty: Teehee... Yes, I don't like it either. It's just that I liked the ten names I thought of before even less. :)

"It builds a `Run`." I'll think about it some more.

tyrian/js/src/main/scala/tyrian/Cmd.scala
uncle-betty: Two commands are equal when they produce equal actions. So, to compare two commands, translate them to their actions via `cmdToTaskList`, and then compare those.

The issue is that `cmdToTaskList` now takes a `Dispatcher` as an argument. So, we now can compare two commands for equality only where there's a `Dispatcher` in scope.

Now that I've slept it over, I think that this may actually kill this PR. The whole point of this PR was to not need a `Dispatcher` in scope in unexpected places.

tyrian/js/src/main/scala/tyrian/runtime/CmdHelper.scala
uncle-betty: Ooooh, interesting! I'll take a look. I'm curious to learn what you changed.

uncle-betty:
Oh, wow! That is an incredibly elegant hat! Awesome!

I was completely unaware of `Resource`'s `allocated`. And I had a hard time trying to make things work solely based on `use`.

I firmly believe that my biggest contribution to the world of open source over the years has been in the form of half-baked PRs that nerd-snipe people who actually know what they're doing. :)

I'm closing this PR, as yours is so much better. Thank you for looking into this!

Just to add some perspective: We're having a fundraising party on Saturday to help our co-working space survive. I used Tyrian to create a simple application that enables the volunteer baristas to track sold drinks. They access the application via tablets and I was worried about the tablets' battery life, because of the current websocket implementation. (We have loooong parties!) It's nice that they will now be able to freely move the tablets around without needing power cords.

Thank you, guys, for helping me with this issue! I appreciate it.

PR #177, opened 2023-03-17
Use FS2 `Channel` for `WebSocket`

armanbilge:
Sorry, I got nerd-sniped 😸 cc @uncle-betty

Like https://github.com/PurpleKingdomGames/tyrian/pull/175, https://github.com/PurpleKingdomGames/tyrian/pull/176, this replaces the busy-loop polling of a mutable queue used to implement `WebSocket`. Instead of using a Cats Effect `Queue`, it uses an FS2 `Channel`, which is optimized for consuming as a `Stream`.

Unlike https://github.com/PurpleKingdomGames/tyrian/pull/176, the `Dispatcher` is not threaded through the runtime. It is kept local to the websocket, so it does not require changes to the runtime.

This PR implements precisely what I said in https://github.com/PurpleKingdomGames/tyrian/pull/175#issuecomment-1468948238: the websocket is a resource, that is _already_ going in the model, so it is no extra effort to handle the `Channel` and `Dispatcher` resources along with it.

PR is best reviewed with whitespace hidden.

### DIFF
--- a/tyrian/js/src/main/scala/tyrian/websocket/WebSocket.scala
+++ b/tyrian/js/src/main/scala/tyrian/websocket/WebSocket.scala
@@ -2,15 +2,16 @@ package tyrian.websocket
 
 import cats.effect.kernel.Async
 import cats.effect.kernel.Sync
-import cats.syntax.functor.*
+import cats.effect.std.Dispatcher
+import cats.syntax.all.*
 import fs2.Stream
+import fs2.concurrent.Channel
 import org.scalajs.dom
 import tyrian.Cmd
 import tyrian.Sub
 import tyrian.websocket.WebSocketEvent
 import util.Functions
 
-import scala.collection.mutable.Queue
 import scala.concurrent.duration.*
 
 /** Helper WebSocket instance to store in your model that acts as a controller */
@@ -18,6 +19,7 @@ final class WebSocket[F[_]: Async](liveSocket: LiveSocket[F]):
   /** Disconnect from this WebSocket */
   def disconnect[Msg]: Cmd[F, Msg] =
     Cmd.SideEffect(liveSocket.socket.close(1000, "Graceful shutdown"))
+      .combine(Cmd.SideEffect(liveSocket.closeChannel))
 
   /** Publish a message to this WebSocket */
   def publish[Msg](message: String): Cmd[F, Msg] =
@@ -29,7 +31,11 @@ final class WebSocket[F[_]: Async](liveSocket: LiveSocket[F]):
     else Sub.emit(f(WebSocketEvent.Error("Connection not ready")))
 
 /** The running instance of the WebSocket */
-final class LiveSocket[F[_]: Async](val socket: dom.WebSocket, val subs: Sub[F, WebSocketEvent])
+final class LiveSocket[F[_]: Async](
+    val socket: dom.WebSocket,
+    val subs: Sub[F, WebSocketEvent],
+    val closeChannel: F[Unit]
+)
 
 enum WebSocketReadyState derives CanEqual:
   case CONNECTING, OPEN, CLOSING, CLOSED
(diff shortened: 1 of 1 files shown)


uncle-betty:
Just tried it and it works like a charm for me. Thanks again!

tyrian/js/src/main/scala/tyrian/websocket/WebSocket.scala
```diff
@@ -2,22 +2,23 @@ package tyrian.websocket
...
 
-import scala.collection.mutable.Queue
 import scala.concurrent.duration.*
 
 /** Helper WebSocket instance to store in your model that acts as a controller */
 final class WebSocket[F[_]: Async](liveSocket: LiveSocket[F]):
   /** Disconnect from this WebSocket */
   def disconnect[Msg]: Cmd[F, Msg] =
-    Cmd.SideEffect(liveSocket.socket.close(1000, "Graceful shutdown"))
+    Cmd.SideEffect(Async[F].delay(liveSocket.socket.close(1000, "Graceful shutdown")) *> liveSocket.closeChannel)
```
gvolpe, line 21: I think this could be `guarantee` instead of `*>`?

tyrian/js/src/main/scala/tyrian/websocket/WebSocket.scala
```diff
@@ -102,59 +107,56 @@ object WebSocket:
...
+                  socket.removeEventListener("message", msgListener)
+                  socket.removeEventListener("error", errListener)
+                  socket.removeEventListener("open", openListener)
+                  socket.removeEventListener("close", closeListener)
+                }
+              ),
+              keepAlive.run
+            )
+
+          LiveSocket(socket, subs, channel.close *> closeDispatcher)
```
gvolpe, line 158: Probably `guarantee` here too, might avoid potential leaks :)

gvolpe (commented):
Code looks good! 👍🏽 I'll give it a try this weekend because the mutable queue was introduced due to some race conditions when the connection is acquired. Want to make sure it still works as expected :)

davesmith00000:
Looks great! Thanks @armanbilge, I'll hang on for @gvolpe who said he wanted to have a look over the weekend, but looks like a good solution, and I see what you meant now. :smile:

tyrian/js/src/main/scala/tyrian/websocket/WebSocket.scala
gvolpe: > if the websocket didn't close, should we close the dispatcher and channel?

I was thinking more about the case where the network call fails, so we don't get the close acknowledgement from the server, but our connection gets closed regardless (think closing a tab). I think the right thing to do is to guarantee the channel is closed as well, after all we are specifically calling `disconnect`, but I see how that can get philosophical 😄

gvolpe:
> Code looks good! 👍🏽 I'll give it a try this weekend 

Done. Can confirm #149 doesn't manifest itself with this version 🚀

tyrian/js/src/main/scala/tyrian/websocket/WebSocket.scala
armanbilge: > I was thinking more about the case where the network call fails

Right, this is an excellent point 🤔 

> I think the right thing to do is to guarantee the channel is closed as well

Can you talk me through exactly how that would work in the case of network failure? I can't see how adding a guarantee here would guard against network failure.

I have a really hard time reasoning about resource safety in Tyrian :)

gvolpe: > I can't see how adding a guarantee here would guard against network failure.

We wouldn't guard against a network failure. We can let it fail while ensuring (via `guarantee`) that the resources are properly closed to avoid potential leaks (channel, and dispatcher that uses allocated).

If that's not convincing enough, feel free to ignore and leave it as is, hardly any free time right now, sorry 😄

tyrian/js/src/main/scala/tyrian/websocket/WebSocket.scala
```diff
@@ -102,59 +107,56 @@ object WebSocket:
...
+            val msg =
+              try e.asInstanceOf[dom.ErrorEvent].message
+              catch { case _: Throwable => "Unknown" }
+            dispatcher.unsafeRunAndForget(channel.send(WebSocketEvent.Error(msg)))
+          }
+
+          val closeListener = Functions.fun { e =>
+            val ev = e.asInstanceOf[dom.CloseEvent]
+            dispatcher.unsafeRunAndForget(channel.send(WebSocketEvent.Close(ev.code, ev.reason)))
+          }
```
armanbilge, line 131: So Gabriel raises a really important issue to make sure that the `Dispatcher` and `Channel` are closed on network error. Since errors are propagated through this callback I think this is where we want to add additional logic. Probably same for the close event listener below as well.

@davesmith00000 would it be safe to assume that once the websocket has errored it is no longer usable?

tyrian/js/src/main/scala/tyrian/websocket/WebSocket.scala
armanbilge: Thanks!

> We can let it fail while ensuring (via `guarantee`) that the resources are properly closed to avoid potential leaks

Yes, absolutely, :100: agree. We need to handle the errors where they are raised, which is actually not here. See https://github.com/PurpleKingdomGames/tyrian/pull/177#discussion_r1141033595.

tyrian/js/src/main/scala/tyrian/websocket/WebSocket.scala
armanbilge: Ah yeah, that seems to be exactly the case.

> The error event is fired when a connection with a WebSocket has been closed due to an error (some data couldn't be sent for example).

https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/error_event

davesmith00000: Sorry I'm late. I know not a lot about WebSockets*, but I read that page the same way you do and agree that closing on error sounds sensible (Which I note you've done in another commit). :+1: 

(*WebSockets have had a lot of attention, but largely but other people! Who knew they'd be so popular?!)

tyrian/js/src/main/scala/tyrian/websocket/WebSocket.scala
davesmith00000: So my reading of this then is that you're going to skip `guarantee` because you've done the 'close on error' work, which seems to be the more correct place to deal with this sort of thing. Have I got that right?

davesmith00000 (approved):
I've tested the branch and it appears to work fine, @armanbilge if you're happy, feel free to merge (or I can).

Thanks! Not only do we have a better solution, I've also learnt some stuff. :smile:

tyrian/js/src/main/scala/tyrian/websocket/WebSocket.scala
armanbilge: Yes, I don't think that `guarantee` makes sense here. From my reading of MDN, `websocket.close()` will only raise an exception if the socket actually didn't close (due to invalid args which should never happen anyway). In which case my opinion is we _don't_ want to close the channel or the dispatcher, if the socket itself is not closed.

https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/close#exceptions

Taking a step back: this actually makes sense. `websocket.close()` is a synchronous call, so it can only raise exceptions for synchronous errors. If there is some sort of network failure, this occurs asynchronously, and would not be raised here.

armanbilge:
Great, thanks for testing that! I don't have merge rights, but if y'all are happy, I'm happy :)

> Not only do we have a better solution, I've also learnt some stuff.

💯 same 😁

gvolpe:
Thanks a lot for digging into it @armanbilge @davesmith00000 ! I was heading out when I commented yesterday, so apologies for the short helpless comment. Still, glad I was the one asking the annoying questions this time to nerd-snipe you a bit more 😆

PR #182, opened 2023-03-23
Fixed #173: Added Cmd.toTask

### LINKED ISSUE #173: Consider adding `Cmd.toTask`
A good question came up on Discord this morning about how you can run a number of `Cmd`s in a strict order, specifically they wanted to run three `Cmd`s in parallel and then a forth based on the result.

Doing this with `Cmd`s themselves is a faff because you have to hold state, but doing this on the underlying `F` type (be that `IO` or `Task` (from `ZIO`) is very easy - it's what they do!

So I wonder if we can have a convenience function that will turn any `Cmd` into a representation in the form of `F` - this is how they are run anyway sooner or later, so it feels like it should be doable. Then you could set up a bunch of `Cmd`s using the standard tooling, convert them to `F`, and build a composite `Cmd.Run` that does exactly what you want.

davesmith00000:
Relates to #173 

The idea is that sometimes you want to build up a computation that starts with an existing `Cmd` and do something with it in terms of it's underlying task. For example you might want to race three HTTP calls, and you can do that by calling `toTask` on the `Cmd`.

Thoughts and feedback welcome from more experienced cat wranglers. :cat:

### DIFF
--- a/tyrian/js/src/main/scala/tyrian/Cmd.scala
+++ b/tyrian/js/src/main/scala/tyrian/Cmd.scala
@@ -50,6 +50,7 @@ object Cmd:
   /** Runs a task that produces no message */
   final case class SideEffect[F[_]](task: F[Unit]) extends Cmd[F, Nothing]:
     def map[OtherMsg](f: Nothing => OtherMsg): SideEffect[F] = this
+    def toTask: F[Unit]                                      = task
   object SideEffect:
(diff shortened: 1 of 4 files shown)


davesmith00000:
(cc @armanbilge)

gvolpe:
Sorry Dave, was at a conference last week, couldn't have a look but looks good 👍🏽

Issue #184, opened 2023-03-27
WebSocket-related exception with 0.6.2-SNAPSHOT


uncle-betty:
Hello there.

I've been playing with 0.6.2-SNAPSHOT as it fixes the increased websocket CPU usage. It works really well! Thank you for that change.

I just noticed one small thing: When I use `disconnect` on a websocket, then there seems to be a race condition in the way things get cleaned up. The resulting `close` message (1000, "Graceful shutdown") reaches the `closeListener` function, but at that point, the `Dispatcher` sometimes has already been released and I get a `dispatcher already shutdown` exception in `closeListener`.

As far as I can tell, this is mostly a cosmetic issue, as the exception simply makes the event listener do nothing. But I thought that I'd bring it up, just in case it has more severe implications that I don't see. I would just have expected the `Dispatcher` to survive until the event listeners had been removed.

It seems to happen more easily on Firefox than on Chrome.

Oh,and I randomly tried turning `channel.close *> closeDispatcher` into `channel.close *> channel.closed *> closeDispatcher`, hoping that `channel.closed` would block on channel shutdown (and postpone `closeDispatcher`) until the event listeners were removed. But that didn't do anything. I guess that I misunderstand at what point the `Sub`'s clean-up code runs (and removes the event listeners).

Again, I'm pointing this out mostly as a curiosity, not as a real problem. I just thought that you might be interested, as this is relatively new code.

uncle-betty:
On second thought - why run `liveSocket.closeChannel` in `disconnect`? Wouldn't `liveSocket.socket.close()` trigger a `close` event and the `closeListener` function would take care of closing the channel and the dispatcher?

Also, isn't an `error` event always followed by a `close` event? So, should `errListener` simply not close anything and instead leave closing things to `closeListener`?

PR #185, opened 2023-03-27
Remove redundant closing in WebSocket implementation

armanbilge:
Hopefully fixes https://github.com/PurpleKingdomGames/tyrian/issues/184.

cc @uncle-betty thanks for investigating this so thoroughly!

Update: because the `'close'` event is guaranteed to be called even in case of error, we can move all on-close handling to there.

### DIFF
--- a/tyrian/js/src/main/scala/tyrian/websocket/WebSocket.scala
+++ b/tyrian/js/src/main/scala/tyrian/websocket/WebSocket.scala
@@ -33,7 +33,7 @@ final class WebSocket[F[_]: Async](liveSocket: LiveSocket[F]):
 /** The running instance of the WebSocket */
 final class LiveSocket[F[_]: Async](
     val socket: dom.WebSocket,
-    val subs: Sub[F, WebSocketEvent],
+    val subs: Sub[F, WebSocketEvent]
 )
 
 enum WebSocketReadyState derives CanEqual:
--- a/tyrian/js/src/main/scala/tyrian/websocket/WebSocket.scala
+++ b/tyrian/js/src/main/scala/tyrian/websocket/WebSocket.scala
@@ -125,17 +125,14 @@ object WebSocket:
             val msg =
               try e.asInstanceOf[dom.ErrorEvent].message
               catch { case _: Throwable => "Unknown" }
-            dispatcher.unsafeRunAndForget(
-              channel.send(WebSocketEvent.Error(msg)) *>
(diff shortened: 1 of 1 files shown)


uncle-betty:
Oh, wow! That was fast. Thank you so much, @armanbilge.

Yes, these changes make the issue go away! Awesome!

Let me try to add to this. Here's my understanding of the fix:

* So, we're exploiting the fact that event listeners for `close` fire in the order in which they are registered.
* This ensures that the `close` event listener registered during `disconnect` runs after `closeListener`.
* So, by the time `disconnect` does `liveSocket.closeChannel`, `closeListener` is finished and we can be certain that we'll never need a `Dispatcher` again.
* So, in particular, when `disconnect` does `liveSocket.closeChannel`, `closeListener` has already done `close.start`.

But isn't that redundant? As far as I can see, `liveSocket.closeChannel` and `close.start` do the same thing, no?

If so, then couldn't we simplify `disconnect` further, by having it run `liveSocket.socket.close()` and nothing else? This would result in a `close` event, which would trigger `closeListener`, which would do `close.start`, which would take care of everything, no?

armanbilge:
> If so, then couldn't we simplify `disconnect` further, by having it run `liveSocket.socket.close()` and nothing else?

Good point! Yes indeed, I think you are right that this is all redundant. So actually we can revert https://github.com/PurpleKingdomGames/tyrian/pull/185/commits/deb63c1be332031a7cce03650b5b16ff12faf466 and just remove `closeChannel` as you say. Which also probably means we don't even need to store it in the model to begin with 🤔

uncle-betty:
Ooooh, right! Yes, good point. `disconnect` seems to be the only user to `closeChannel`. So, no need to keep it around now.

uncle-betty:
I just tested the latest version and it works like a charm for me as well. Tested on Firefox, Chrome, and Mobile Safari.

davesmith00000 (approved):
Lovely work!

davesmith00000:
Thanks @uncle-betty and @armanbilge. Great work. :pray: 

I've reviewed the changes, read the through discussion, run the tests, and done some superficial manual testing (confident @uncle-betty testing will have been more thorough!). Anything else to do? Have I missed anything, or are you both happy for the changes to be merged?

armanbilge:
Thanks all, I'm happy, this was a very nice simplification over the original fix :)

PR #187, opened 2023-03-30
Issue/178/bug property types

### LINKED ISSUE #178: Bug: Support more Types for Properties
Currently the `Property` case class only supports `String` values, however this does not work in cases such as Booleans, i.e. `checked = true` is not the same as `checked = 'true'`.

The underlying snabbdom lib accepts `Any` type, so we just need to enrich the API a little to support a few basic types:

- `Double`
- `Int` (not strictly necessary but...)
- `String`
- `Boolean`
- JS Functions. (`Functions.fun1` for example?)

The only slight complication to this is that it affects the code gen stuff that happens at build time.

davesmith00000:
Fixes #178.

The specific case we had involved (in HTML) `checked`, like this:

`<input type="checkbox" checked>foo</input>`

Where I mistakenly thought that modifying the presence of `checked` would toggle the checkbox on and off. Not so, `checked` as used above is an attribute that defines the state of the checkbox _on page load_ but not after.

In order to programmatically check the box after page load you need to modify the _property_ `checked`, not the attribute. It's to do with [IDL](https://developer.mozilla.org/en-US/docs/Glossary/IDL) and how that reflects the value of the attributes, or not.  Luckily there do not appear to be many attributes where it's really necessary to define a corresponding property. The list is:

- value (String)
- selected (Boolean)
- checked (Boolean)
- indeterminate (Boolean) - when a checkbox is neither selected or unselected for example.

There are others like `defaultChecked`, but that seems irrelevant in a VirtualDom scenario, since we already know whether or not an input was checked at page load time.

So this PR adds typed properties for those mentioned, and also now restricts `value` to being just a `String`. `value` always was a string really but we had convenience methods to help you use other types.

In the case of `checked`, you can now do this:

```scala
checked // attribute - is checked at page load
checked(true|false) // attribute - is/isn't checked at page load
checked := true|false // property - checks and unchecks at any time.
```

### DIFF
--- a/project/AttributeGen.scala
+++ b/project/AttributeGen.scala
@@ -11,10 +11,13 @@ object AttributeGen {
           else "value.toString"
         })
       |
-      |  final class PropertyName$typ(name: String):
-      |    def :=(value: $typ): Property = Property(name.toString, ${
-          if (typ == "String") "value" else "value.toString"
-        })
+      |""".stripMargin
+    }.mkString
+
+  def generatePropertyNameTypes: String =
+    List("String", "Boolean").map { typ =>
+      s"""  final class PropertyName$typ(name: String):
+      |    def :=(value: $typ): Property$typ = Property$typ(name.toString, value)
       |
       |""".stripMargin
     }.mkString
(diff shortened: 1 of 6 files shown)


davesmith00000:
Dear people-I've-tagged-as-reviewers,

I mostly raise PR's for visibility, record keeping, and so I can look them over with fresh eyes in the morning. Reviews are always appreciated, but please don't stress over it.

This is supposed to be fun. :smile:

PR #188, opened 2023-04-01
Fixed #110: Event.preventDefault support

### LINKED ISSUE #110: Consider invoking preventDefault on a form's button
Given a following form:

```html
<form>
  <input type=text name=email />
  <input type=password name=password />
  <button onclick=submit()>Login</button>
</form>
```

It must be quite common to do [`e.preventDefault()`](https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault) as first thing in `submit` callback, because otherwise clicking on the button makes the browser to send a form and reload current page.

It's possible in plain JS (and had to do in scalajs-react), but Tyrian doesn't allow me to call it, so I had to revert to `a href=#`. However, I believe it must be a defult behaviout for any SPA as we don't want to reload page.

davesmith00000:
Resolves #110.

Events now `preventDefault`, `stopPropagation`, and `stopImmediatePropagation` by default, because have read around the subject, this seems like the behaviour we want in an Elm-like application.

I have expanded the API so that you can do things like `preventDefaultOn`, `preventDefaultOff`, `preventDefaultToggle`, in case you want the standard behaviour:

```scala
form(onSubmit(Msg.Log("submitted").preventDefaultOff))(
  input(_type := "submit", name := "submit", value := "submit")
)
```

I don't know if this is the most elegant solution, but it gets around the immediate problems and provides and more sensible default. We can always improve on this later.

### DIFF
--- a/tyrian/shared/src/main/scala/tyrian/Attr.scala
+++ b/tyrian/shared/src/main/scala/tyrian/Attr.scala
@@ -93,16 +93,20 @@ final case class Event[E <: Tyrian.Event, M](
     stopPropagation: Boolean,
     stopImmediatePropagation: Boolean
 ) extends Attr[M]:
-  def map[N](f: M => N): Attr[N]                  = Event(name, msg andThen f)
-  def preventDefaultToggle: Event[E, M]           = this.copy(preventDefault = !preventDefault)
-  def preventDefaultOn: Event[E, M]               = this.copy(preventDefault = true)
-  def preventDefaultOff: Event[E, M]              = this.copy(preventDefault = false)
-  def stopPropagationToggle: Event[E, M]          = this.copy(stopPropagation = !stopPropagation)
-  def stopPropagationOn: Event[E, M]              = this.copy(stopPropagation = true)
-  def stopPropagationOff: Event[E, M]             = this.copy(stopPropagation = false)
-  def stopImmediatePropagationToggle: Event[E, M] = this.copy(stopImmediatePropagation = !stopImmediatePropagation)
-  def stopImmediatePropagationOn: Event[E, M]     = this.copy(stopImmediatePropagation = true)
-  def stopImmediatePropagationOff: Event[E, M]    = this.copy(stopImmediatePropagation = false)
+  def map[N](f: M => N): Attr[N] = Event(name, msg andThen f)
+
+  def withPreventDefault(enabled: Boolean): Event[E, M] = this.copy(preventDefault = enabled)
+  def usePreventDefault: Event[E, M]                    = withPreventDefault(true)
+  def noPreventDefault: Event[E, M]                     = withPreventDefault(false)
+
+  def withStopPropagation(enabled: Boolean): Event[E, M] = this.copy(stopPropagation = enabled)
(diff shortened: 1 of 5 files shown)


davesmith00000:
cc @chuwy

tyrian/shared/src/main/scala/tyrian/Attr.scala
```diff
@@ -86,5 +86,29 @@ object Property:
...
-  def map[N](f: M => N): Attr[N] = Event(name, msg andThen f)
+final case class Event[E <: Tyrian.Event, M](
+    name: String,
+    msg: E => M,
+    preventDefault: Boolean,
+    stopPropagation: Boolean,
+    stopImmediatePropagation: Boolean
+) extends Attr[M]:
+  def map[N](f: M => N): Attr[N]                  = Event(name, msg andThen f)
+  def preventDefaultToggle: Event[E, M]           = this.copy(preventDefault = !preventDefault)
```
hobnob, line 97: I can't think of a use case for this, did you have something in mind?

tyrian/shared/src/main/scala/tyrian/Attr.scala
```diff
@@ -86,5 +86,29 @@ object Property:
...
+  def preventDefaultOn: Event[E, M]               = this.copy(preventDefault = true)
```
hobnob, line 98: Is it worth using `withPreventDefault`/`withoutPreventDefault`?

tyrian/shared/src/main/scala/tyrian/Attr.scala
```diff
@@ -86,5 +86,29 @@ object Property:
...
+  def preventDefaultOff: Event[E, M]              = this.copy(preventDefault = false)
+  def stopPropagationToggle: Event[E, M]          = this.copy(stopPropagation = !stopPropagation)
+  def stopPropagationOn: Event[E, M]              = this.copy(stopPropagation = true)
```
hobnob, line 101: Likewise for these ... just brings it inline with Indigo really, but happy to go either way :)

hobnob (approved):
Just the naming thing really, but I won't be offended if you ignore me ;)

tyrian/shared/src/main/scala/tyrian/Attr.scala
davesmith00000: Having stewed on it a bit I think you're right. I can't think of a use-case either. :smile: :+1:

tyrian/shared/src/main/scala/tyrian/Attr.scala
davesmith00000: Is Indigo's approach the best - don't know, but bringing it inline isn't a bad shout. That would mean this sort of thing:

```
  def withTransparentBackground(enabled: Boolean): GameConfig =
    this.copy(transparentBackground = enabled)
  def useTransparentBackground: GameConfig =
    withTransparentBackground(true)
  def noTransparentBackground: GameConfig =
    withTransparentBackground(false)
```

`with`/`use`/`no`.

I can't decide whether one is better than the other, but I'll bow to consistency.

Issue #193, opened 2023-04-16
Export CSS properties in camel case too


davesmith00000:
Currently CSS (under import Tyrian.CSS.*) properties are exported in kebab case, but some IDE's appear to struggle with that, so it would be good to also export them as camel case where no backticks are required.

davesmith00000:
This was done, will be in the next release.

PR #194, opened 2023-04-16
#193 CSS properties in camel-case

krined-dev:
I'm not sure if this is the correct way to go about this or if there is an easier way to alias imports that don't require the function to be redefined. 

example prop: animation-delay
output
```
  def `animation-delay`(value: String): Style = Style("animation-delay", value)
  def animationDelay(value: String): Style = Style("animation-delay", value)
```

### DIFF
--- a/project/CSSGen.scala
+++ b/project/CSSGen.scala
@@ -3,8 +3,22 @@ import scala.sys.process._
 
 object CSSGen {
 
-  def genCssProp(name: String): String =
-    s"""  def `$name`(value: String): Style = Style("$name", value)"""
+  def genCssProp(name: String): String = {
+    val cssProp = s"""  def `$name`(value: String): Style = Style("$name", value)"""
+    if (!name.contains("-") || name.contains("@")) { 
+      cssProp
+    } else {
+      val nameCC: String = name.split("-").toList match {
+        case Nil =>
+  	  throw new Exception(s"CSS property '$name' was unexpectedly empty")
+        case head :: tail =>
+  	  (head :: tail.map(_.capitalize)).mkString
+      }
+      s"""$cssProp 
+      |  def ${nameCC}(value: String): Style = Style("$name", value)
+      |""".stripMargin
+    }
+  }
 
   def template(moduleName: String, fullyQualifiedPath: String, contents: String): String =
     s"""package $fullyQualifiedPath


krined-dev:
This does not deal with properties that contain @ as well. I don't think there's anything to do about them.

davesmith00000:
This is a good attempt and I'm delighted to have the contribution. I am going to leave some comments for you to have a look at though. :smile:

project/CSSGen.scala
```diff
@@ -3,8 +3,16 @@ import scala.sys.process._
 
 object CSSGen {
 
-  def genCssProp(name: String): String =
-    s"""  def `$name`(value: String): Style = Style("$name", value)"""
+  def genCssProp(name: String): String = {
+    val prop = s"""  def `$name`(value: String): Style = Style("$name", value)"""
+    if (!prop.contains("-") || prop.contains('@')) { prop }
```
davesmith00000, line 8: I think you want to be testing `name` in the predicate, not `prop`, so:
`if (!name.contains("-") || name.contains('@')) { prop }`

However, you also don't need the `!name.contains("-")` test, because spliting a string without a hyphen is just a list of one element, and should work (I think).

project/CSSGen.scala
```diff
@@ -3,8 +3,16 @@ import scala.sys.process._
...
+    else {
+      val namePC: String = name.split("-").map(_.capitalize).mkString
```
davesmith00000, line 10: Here you've got yourself into trouble early by capitalising everything, and you try and fix it later by doing a head + tail on the string.

If might work! But there's a cleaner way, which incidentally, also appears to work for properties beginning with `@`:

```scala
val name = "@my-fancy-prop"
name.split("-").toList match {
  case Nil =>
  	throw new Exception(s"CSS property '$name' was unexpectedly empty")

  case head :: tail =>
  	(head :: tail.map(_.capitalize)).mkString
}
```

project/CSSGen.scala
```diff
@@ -3,8 +3,16 @@ import scala.sys.process._
...
+      (s"""  def `$name`(value: String): Style = Style("$name", value)""",
```
davesmith00000: This line is a duplicate of `prop`, so you can replace it with the variable.

project/CSSGen.scala
krined-dev: Yeah, name was definitely what I meant there. checking if it contains the hyphen is just to avoid double declaration of the method.

project/CSSGen.scala
krined-dev: Yeah, that is a lot cleaner :) @ will have to be backticked anyway right? Because you can't declare a function in scala starting with @.

project/CSSGen.scala
```diff
@@ -3,8 +3,16 @@ import scala.sys.process._
...
+       s"""  def ${s"${namePC.head.toLower}${namePC.tail}"}(value: String): Style = Style("$name", value)""")
+      .productIterator.mkString("\n")
```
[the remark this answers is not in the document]
krined-dev: I think it should just be string concatenated though, my brain is just slightly broken after exam prepping :D

project/CSSGen.scala
davesmith00000: Ah! Yes sorry that is a excellent point. :+1:

project/CSSGen.scala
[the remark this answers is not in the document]
davesmith00000: You could even have them both in the same block, off the top of my head:

```scala
s"""  def `$name`(value: String): Style = Style("$name", value)"""
|  def ${s"${namePC.head.toLower}${namePC.tail}"}(value: String): Style = Style("$name", value)
|""".stripMargin
```

project/CSSGen.scala
krined-dev: Might be something to mention in the README/docs? I'm not sure if it's a problem specific enough to Tyrian to mention though, but could have a disclaimer like "Most properties have a camel case alias, except for properties that begin with @".

davesmith00000 (approved):
Look great! Thank you very much for the work. :+1:

PR #197, opened 2023-05-02
Frontend Routing

davesmith00000:
The aim of this PR is to build frontend routing into Tyrian.

The plan is to do the same as Indigo, i.e. depending on which initial trait you extend you get different behaviour. In this case, `SinglePage` is the replacement for `TyrianApp`, and `MultiPage` will give you almost the same API, but plumbs the routing into your app for you.


davesmith00000:
Build is failing on unrelated test cases...

davesmith00000:
Progress so far: I've got the two entry points defined, and I've simply lifted the existing `Navigation` module (which I have not attempted to improve) to provide location hash based routing just to prove it works.

Ugly, basic, but working.

The next thing will be to replace the hash based navigation with something more general purpose, and to clean up the API so that the requirement is for you to provide a simple function like this:

```scala
def router: Location => Msg 
```

tyrian/js/src/main/scala/tyrian/TyrianRoutedAppF.scala
```diff
@@ -0,0 +1,45 @@
...
+  */
+trait TyrianRoutedAppF[F[_]: Async, Msg, Model] extends TyrianAppF[F, Msg, Model]:
+
+  def hashResultToMessage: Navigation.Result => Msg
+  def hashChangeToMessage: Navigation.Result.HashChange => Msg
+
+  def setLocationHash(newHash: String): Cmd[F, Nothing] =
+    Navigation.setLocationHash(newHash)
+
+  def _init(flags: Map[String, String]): (Model, Cmd[F, Msg]) =
```
bilki, line 22: Perhaps you might want to make all these `_`-prefixed methods private? Otherwise they'll leak into `SinglePage` and `MultiPage`, is that on purpose? 🤔

sandbox/src/main/scala/example/Sandbox.scala
```diff
@@ -17,7 +17,14 @@ import scala.scalajs.js.annotation.*
 import scalajs.js
 
 @JSExportTopLevel("TyrianApp")
-object Sandbox extends TyrianApp[Msg, Model]:
+object Sandbox extends MultiPage[Msg, Model]:
+
+  val hashResultToMessage: Navigation.Result => Msg = {
+    case Navigation.Result.CurrentHash(hash) => Msg.NavigateTo(Page.fromString(hash))
+    case _                                   => Msg.NavigateTo(Page.Page1)
```
bilki, line 24: 🤔 wondering if there's a way to bake those `NavigateTo` messages into `Msg` generic type somehow. They look like they'll be always used when dealing with a `MultiPage`, so why not enforce/facilitate them? Probably not a topic for this PR but the next one generalizing the router.

tyrian/js/src/main/scala/tyrian/TyrianRoutedAppF.scala
```diff
@@ -0,0 +1,45 @@
...
+/** The TyrianApp trait can be extended to conveniently prompt you for all the methods needed for a Tyrian app, as well
+  * as providing a number of standard app launching methods.
+  */
+trait TyrianRoutedAppF[F[_]: Async, Msg, Model] extends TyrianAppF[F, Msg, Model]:
+
+  def hashResultToMessage: Navigation.Result => Msg
+  def hashChangeToMessage: Navigation.Result.HashChange => Msg
+
+  def setLocationHash(newHash: String): Cmd[F, Nothing] =
+    Navigation.setLocationHash(newHash)
```
bilki, line 20: Is it worth enough adding this method here? I don't see any advantages to using it compared to calling `Navigation.setLocationHash` directly (and users already need to import `Navigation.Result` and others to implement hash result and change to message functions)

tyrian/js/src/main/scala/tyrian/TyrianRoutedAppF.scala
```diff
@@ -0,0 +1,45 @@
...
+
+import scala.scalajs.js.annotation._
+
+/** The TyrianApp trait can be extended to conveniently prompt you for all the methods needed for a Tyrian app, as well
+  * as providing a number of standard app launching methods.
+  */
+trait TyrianRoutedAppF[F[_]: Async, Msg, Model] extends TyrianAppF[F, Msg, Model]:
+
+  def hashResultToMessage: Navigation.Result => Msg
+  def hashChangeToMessage: Navigation.Result.HashChange => Msg
```
bilki, line 17: I'd rather make it clear to the user that these functions are used to actually spit a message **everytime** the app initializes and the hash/URL changes. Maybe it's worth adding a paragraph in a new docs section for `Routing`. Anyway, they'll be forced to deal with the messages in the `update` function so not critical in the end.

indigo-sandbox/src/main/scala/example/IndigoSandbox.scala
```diff
@@ -12,7 +13,7 @@ import zio.interop.catz.*
 import scala.scalajs.js.annotation.*
 
 @JSExportTopLevel("TyrianApp")
-object IndigoSandbox extends TyrianApp[Msg, Model]:
+object IndigoSandbox extends SinglePage[Msg, Model]:
```
bilki, line 16: I don't think it's a huge concern since `TyrianApp` is only used here and people bumping Tyrian version would need to change a single line. But it might be worth to keep this type and make it a type alias (or extend from `SinglePage`), what do you think?

bilki (commented):
Looking good, just added a few comments 😄

tyrian/js/src/main/scala/tyrian/TyrianRoutedAppF.scala
davesmith00000: Unsure, quite honestly. Mostly people will change the address by clicking a link, this is just here so you can do the same thing from a model update. I was sort of thinking "if it's hear, they don't need to know about `Navigation`, but it feels weird. Really all we're doing is setting `window.location.hash`, and I could just leave people to do that themselves. Again, deferring that decision to later.

tyrian/js/src/main/scala/tyrian/TyrianRoutedAppF.scala
davesmith00000: I'm hoping these two go away / are replaced. They're just a hangover from `Navigation`. :crossed_fingers:

indigo-sandbox/src/main/scala/example/IndigoSandbox.scala
davesmith00000: I'm a bit torn. I'm kinda unafraid of forcing people to keep up in general, but I could put in a type alias with a deprecation warning if I was feeling kind... :smile:

davesmith00000:
You may see failing builds here, but it's a flaky test... https://github.com/PurpleKingdomGames/tyrian/issues/199

indigo-sandbox/src/main/scala/example/IndigoSandbox.scala
davesmith00000: I'll do the depreciation warning. Seems the right thing to do.

bilki (approved):
Looks good, nice as a starting point.

tyrian/js/src/main/scala/tyrian/TyrianRoutedAppF.scala
```diff
@@ -0,0 +1,75 @@
...
+        _subscriptions,
+        MaxConcurrentTasks
+      )
+    )
+
+object Routing:
+
+  private def locationToRoute(loc: Location): String =
+    val origin = loc.origin.getOrElse("")
+    loc.toString.replaceFirst(origin, "")
```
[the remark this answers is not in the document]
davesmith00000: Ya know when I read the description on that originally I thought it did something else, but I've double checked and tried and you are 100% correct. :bow:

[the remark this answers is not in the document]
davesmith00000: I spoke too soon! When I said I'd tried it, I tried it on a random website in the JS console. In the sandbox it doesn't work because it gives you the path but omits hash locations. I'll figure something out. :thinking:

sandbox/src/main/scala/example/Sandbox.scala
davesmith00000: So I think at the moment I'm not going to worry about this. The router requires you to decide what to do when a route changes by producing a `Msg`. As you say, maybe follow up in another PR if there turns out to be a common solution, but I don't want to second guess one at the moment.

tyrian/js/src/main/scala/tyrian/TyrianRoutedAppF.scala
davesmith00000: I have done this in the end, it's now `Routing.setLocation`. (`Navigation` has been removed.)

indigo-sandbox/src/main/scala/example/IndigoSandbox.scala
davesmith00000:
```
trait TyrianApp in package tyrian is deprecated: Please use SinglePage or MultiPage instead of TyrianApp.
```

davesmith00000:
@hobnob & @bilki: I've resolved the old discussion points, either by fixing them or because they're out of date - hopefully I didn't missing anything.

(cc @DeamonDev in case you have any thoughts on this. Note that I'm not planning to bake any third party url parsing libs into Tyrian directly, but it would be ideal if this solution optionally worked with libs like `url-dsl`, as you're doing in #195.)

This all works, but it needs more thought.

## Problem 1 - Hyperlinks
The first (and maybe main) problem is that you need to set hyperlinks like this at the moment:

```scala
a(href := "#", onClick(Msg.NavigateTo(page)))("my link")
```

Rather than the more desirable (I think? Probably better for accessibility too?):

```scala
a(href := "/my-page")("my link")
```

To make the latter work, I think we'd have to somehow set `preventDefault` on all `a` tags, effectively:

```html
<a href="/my-page" onclick="event.preventDefault()">my link</a>
```

The catch with that is that I think I'd have to patch it into the runtime somewhere as a special case. :thinking: 

## Problem 2 - Routing UX
The second problem is the routing function. I thought this was all going smoothly until I realised just how nuanced routing can potentially get. :sweat_smile: 

At the moment I've dodged the problem by introducing a `Location` type and basically making all routing decisions the users problem. Want to work on hash changes? Check the hash field. Want to use full paths? Carry on. ...but I feel like we could do better in terms of user experience somehow...

Do we need to do that now? Could it be a future enhancement based on user feedback? Looking up how other frameworks do this is a bit of a minefield. Recommended reading suggestions welcome. 

## Problem 3 ...what have I missed?
I'm super confident that there is more complexity here I'm missing because I'm sort of building this in abstract without a concrete use case. Suggestions welcome.

hobnob:
Looks good Dave!

You're right that the second option for `a` tags would be more accessible, and user friendly. Also means that you don't break how links work (for example the user can ctrl+click to open a link in a new tab).

The question of how to achieve it though is a little tricky 🤔 I wonder if we could modify the `a` so that it took a new attribute called `route`. You could then write: `a(route := Route('/page/{1}/{2}', 'someParam', 'someParam2'))("My text")` and Tyrian would render it with the correct markup (in this case `<a href="/page/someParam/someParam2" onclick="event.preventDefault()">My text</a>`)

Additionally, you could also create a new `Route` object that takes a number of additional parameters, such as base URL and route parameters. A simple implementation might look like this (I haven't compiled it, so it's probably wrong):
```scala
final case class Route(baseUrl: String, path: String, routeParams: Seq[String]) {
}

object Route:
  def apply(path: String, routeParams: Seq[String]): Route =  Route("", path, routeParams)
  def apply(path: String): Route = Route("", path, [])
```

The benefit here would be that you could do a check against a string (such as a URL) against a route, and could also use a base URL in case you were using Tyrian in a subfolder on a main site (such as a blog) without having to change loads of code if the Base URL changes.

Hope that helps 🙂

davesmith00000:
Thanks @hobnob - I've mulled it over and it's a good idea as a way to give us a hook to decide what to do next, I will need something like that. I would like to avoid adding a non-standard entity into the markup however (if possible). Ideally if you can HTML, you should be able to Tyrian (and if you can't HTML, you can google HTML and apply almost 1:1). (Have we already added non-standard stuff with the events? Maybe? Gah! :thinking:)

As usual, all roads lead to Elm. I've had a bit of a snoop around what they do (I haven't read the code, just mooched around the docs to borrow wisdom) and if applied directly, I think it goes something like this:

1. Intercept `a` tags with an `href`.
2. If the tag has an `onClick`, do nothing, our programmer has a plan (and `preventDefault` will happen by default).
3. If there is no `onClick` then the user intends a normal link, so we add an `onClick` that:
  1. Applies `preventDefault`.
  2. Captures the link and turns it into a type similar to the new `Location` type, but that is an enum of `Internal` and `External`.
  3. Note: There is a problem here to overcome in so much as we don't have access to the user's `Msg` type, so we may need to wrap it up or something.
4. The data is then passed to the router, which now runs _before_ any action is taken on the link.
5. The router is modified so that the signature is something like `def router: LocationEnum => NavTo[Msg]`
6. This effectively captures all unhanded links in the App, gives them to the user, and lets them decide whether to just follow them or to modified their app based on the route.
7. The `NavTo` type above is then use to decide what sort of `pushstate` / location update to do.
8. Open question: What happens with the `Sub` in this scenario? I think it'll come out in the wash and we'll still need it (in case a link is entered manually for example) but it's relationship with the router is a little unclear. However it may be that we don't reorder when the router is run though (point (4)) and then it's all ok, we'll see how it pans out.

This isn't a perfect translation of the Elm solution, but I think this is the broad strokes of what the shape of the solution needs to be. No doubt I'll discover more problems along the way.

On reflection, I think some of this goes towards a suggestion @bilki had in an earlier comment around having some generic nav types.

tyrian/js/src/main/scala/tyrian/TyrianRoutedAppF.scala
```diff
@@ -0,0 +1,161 @@
...
+      _hash = "",
+      _protocol = "",
+      _search = "",
+      _href = "",
+      _hostname = "",
+      _port = "",
+      _pathname = path,
+      _host = "",
+      _origin = None
+    )
```
davesmith00000, line 147: Needs a full test suite and ~much better~ actual path parsing.

davesmith00000:
This version has a few warts, but appears to work correctly. There's a couple of TODOs to address (called out in comments), and the only other thing is that the initial routing has to happen last, by which time we've briefly (like a flicker) rendered the homepage even if we want some other page.

davesmith00000:
Getting there, two outstanding issues that I know about:

1. Following anchor/hash links isn't working in the sandbox, this is because you just get the path from the anchor onwards. Just need to decide how to handle that (and whether it's a user problem or an engine problem.

2. Following external links is producing an error: `Uncaught DOMException: The operation is insecure.`

davesmith00000:
I think I'm done with this now. The end result is that `TyrianApp` is back and everyone has to do _some_ routing. However I've made it pretty easy (I think).

Your Tyrian web app now need to implement `def router: Location => Msg`, and there are a few ways to do that (note that in all cases, you're told if the link is considered internal or external):

1. You can use the helpers if your needs are simple, for example, this one ignores all internal links and just routes external ones. Note that you need to provide suitable msg's:

```scala
def router: Location => Msg = Routing.externalOnly(Msg.NoOp, Msg.FollowLink(_))
```

I've also make `Routing.basic` (simple string match) and `Routing.none` variations.

2. Alternatively, you can do a full match, like this:

```scala
  def router: Location => Msg =
    case loc: Location.Internal =>
      loc.pathName match
        case "/"      => Msg.NavigateTo(Page.Page1)
        case "/page1" => Msg.NavigateTo(Page.Page1)
        case "/page2" => Msg.NavigateTo(Page.Page2)
        case "/page3" => Msg.NavigateTo(Page.Page3)
        case "/page4" => Msg.NavigateTo(Page.Page4)
        case "/page5" => Msg.NavigateTo(Page.Page5)
        case "/page6" => Msg.NavigateTo(Page.Page6)
        case _        => Msg.NoOp

    case loc: Location.External =>
      Msg.NavigateToUrl(loc.href)
```

The `Location` type comes with all sorts of information in it for you to decide how to route, it is similar to the JavaScript `Location` type.

I imagine that you could also decide to pull in some [fancy url parsing library](https://github.com/sherpal/url-dsl) to build a clever router here if you liked.

One you've done your routing, you'll want to use some of the new `Nav` cmds. For internal links, you can use `Nav.pushUrl` to update the address bar, and for external, you can use `Nav.loadUrl` to tell the browser to follow the link.

Issue #198, opened 2023-05-04
Proposal: Generic launch example


stevechy:
I was trying out Tyrian and wrote a simple launcher based on how [react-rails](https://github.com/reactjs/react-rails) mounts elements.

There's some extra Promise wrapping around it but the basic idea is.

```
        for {element <- dom.document.querySelectorAll("[data-tyrian-app]")
             } yield {
          val htmlElement = element.asInstanceOf[HTMLElement]
          htmlElement.dataset.get("tyrianApp") match {
            case Some("MyCustomApp") => {
              println("Launching")
              new MyCustomApp().launch(element)
            }
            case _ => ()
          }
        }
```

Flags could also be read from the dataset

Does this seem useful?

davesmith00000:
Yes I like it. So if I'm reading it right, the idea would be to launch from the main method, and that essentially does the above? (+ flags in the dataset)

Have I got that right?

One of the common questions I get quite often is "why can't I just use the main method?" and the answer is of course that you can! ...but we don't have many examples, nor is there a good built in solution for that.

stevechy:
Yep, I'm currently launching this from the main method.  It looks like if the bundle is loaded as a regular script vs module it has to wait for DOMContentLoaded so I also have some handling for that.

I could create a PR with an example in the examples folder?

davesmith00000:
Hi @stevechy - I was wondering about baking it into the main launcher, here somewhere:

https://github.com/PurpleKingdomGames/tyrian/blob/main/tyrian/js/src/main/scala/tyrian/TyrianAppF.scala

Although I confess that at the moment I don't know exactly what that would look like!

If you felt like having a go, you could either work on the main launcher, or add an example - as you suggested - that we could use as the basis for the launcher version later (that I could help with). It would be great to have either contributed if you have the time, no pressure. :smile:

stevechy:
Ok, I created https://github.com/PurpleKingdomGames/tyrian/pull/200 I guess I should have made it a draft.

stevechy:
Having multiple apps probably doesn't fit with the Elm Architecture but I find it useful to get things up and running.

```
  <div data-tyrian-app="CounterApp" data-tyrian-app-flag-initial-counter="42"></div>

  <div data-tyrian-app="ChatApp" data-tyrian-app-flag-initial-message="Hello"></div>
```

davesmith00000:
[...]

Issue #199, opened 2023-05-05
Flakey http tests `tyrian.HttpTests`


davesmith00000:
They are pretty slow and fail intermittently. Here for example the HEAD test passed, but I've seen it fail too, despite no code changes in that area.

```
tyrian.HttpTests:
  + HEAD 1.33s
  + POST 1.09s
  + PUT 3.28s
  + PATCH 0.40s
  + DELETE 1.65s
==> X tyrian.HttpTests.GET with headers 0.93s munit.ComparisonFailException: 
```

davesmith00000:
Annoyingly I can't replicate this now. Will leave the issue open a bit longer...

JPonte:
Seems like the tests I added 😞 I wonder if httpbin was having issues at the time? That's the downside of depending on an external service for the unit tests...

davesmith00000:
Yeah it's being flaky again this morning - no worries though, it's still a massive improvement to have tests for this stuff!

I think maybe I'll just have to switch httpbin out for some sort of mock when I get a chance.

davesmith00000:
I've disable the tests for now and left the following comment in the code while I have a think about what to do next:

```
    // These tests work great, but http://httpbin.org is unreliable, it seems.
    // So we either need an HTTP mock server compatible with Scala.js, or to use
    // containers (I'd rather not, but...). In the meantime these failing tests
    // are lowering build confidence, so I've disabled them.
```

I do want to bring them back though! Just not quite sure how best to proceed. httpbin is available as a Docker image, but I really dislike docker. :yum:

JPonte:
Understandable, I also thought about some sbt magic that would start the httpbin container just for the tests but I didn't like the idea and my sbt knowledge is also questionable. I think @armanbilge had some fancy setup in their project with some simple http4s server running only for the tests or something? Maybe we can make an http4sbin 😅

davesmith00000:
> http4sbin

The thought crossed my mind... 😄

davesmith00000:
I happened to be manually testing an http call to httpbin this morning and it was a bit slow. I wonder if we could get away with just having retries with a backoff policy...

JPonte:
Were you thinking of adding the retry policy to the API itself or just for the tests?

PR #200, opened 2023-05-05
Main method launcher #198

stevechy:
Implementing a launch method to use in `main()`

https://github.com/PurpleKingdomGames/tyrian/issues/198


tyrian/js/src/main/scala/tyrian/TyrianAppF.scala
```diff
@@ -96,3 +99,47 @@ trait TyrianAppF[F[_]: Async, Msg, Model]:
 
       case None =>
         throw new Exception(s"Missing Element! Could not find an element with id '$containerId' on the page.")
+
+object TyrianAppF:
+  /** Launch app instances after DOMContentLoaded.
+    */
+  def launchOnContentLoaded[F[_] : Async](appDirectory: Map[String, () => TyrianAppF[F, _, _]]): Unit = {
+    val documentReady = new js.Promise((resolve, _reject) => {
```
stevechy, line 107: If the main `<script>` tag uses `module` or `defer` then it looks like the DOM will be ready when the script is loaded. In that case you could call `launch` directly.

```
<script type="module" src="/index.hash1.js"></script>
<script src="/index.hash2.js" nomodule defer></script>
```

tyrian/js/src/main/scala/tyrian/TyrianAppF.scala
```diff
@@ -96,3 +99,47 @@ trait TyrianAppF[F[_]: Async, Msg, Model]:
...
+          // Log a warning message?
+          ()
+    }
+  }
+
+  private def appElementFlags(tyrianAppElement: dom.HTMLElement): Map[String,String] =
+    val appFlags = for {
+      (dataAttr, attrValue) <- tyrianAppElement.dataset
+      if dataAttr.startsWith("tyrianAppFlag")
+      flagName = dataAttr.replaceFirst("^tyrianAppFlag", "")
```
stevechy, line 143: This makes the flag names UpperCamelCase

examples/main-launcher/src/main/scala/example/Main.scala
```diff
@@ -0,0 +1,14 @@
+package example
+
+import cats.effect.IO
+import tyrian.Html.*
+import tyrian.*
+
+object Main {
+  def main(args: Array[String]): Unit =
+    TyrianAppF.launchOnContentLoaded(Map(
```
davesmith00000, line 9: This needs to change slightly, people shouldn't be pulling in `TyrianAppF`, they should be using `TyrianApp` from the `IO` or `ZIO` modules. Maybe we just add a companion object there that delegates to this thing?

davesmith00000: I think also we could find a different name here to `launchOnContentLoaded`. :thinking: 

Is  `TyrianApp.onLoad(Map(..))` descriptive enough?

Would also be good to allow another constructor:

`def launchOnContentLoaded[F[_] : Async](appDirectory: (String, () => TyrianAppF[F, _, _]])*): Unit`

So that people can omit the `Map`:

```scala
TyrianAppF.launchOnContentLoaded(
  "CounterApp" -> () => CounterApp,
  "ChatApp" -> () => ChatApp
)
```

davesmith00000: I wonder if it's also possible to leave out the `thunk` syntax somehow? So you just get:

```scala
TyrianAppF.launchOnContentLoaded(
  "CounterApp" -> CounterApp,
  "ChatApp" -> ChatApp
)
```

What do you think?

davesmith00000:
I've approved and run the workflow - please note that we have a couple of flaky http tests at the moment, so if it fails there it isn't you! :)

https://github.com/PurpleKingdomGames/tyrian/issues/199

examples/main-launcher/src/main/scala/example/Main.scala
```diff
@@ -0,0 +1,14 @@
...
+      "CounterApp" -> (() => CounterApp),
+      "ChatApp" -> (() => ChatApp)
+    ))
+    println("")
```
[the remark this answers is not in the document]
stevechy: Oh thanks, this it was leftover from debugging, I will remove it

examples/main-launcher/src/main/scala/example/Main.scala
stevechy: Ok, I have added the companion objects, changed the method name to `onLoad` with varargs and removed the thunk syntax.

tyrian/js/src/main/scala/tyrian/TyrianAppF.scala
```diff
@@ -96,3 +99,47 @@ trait TyrianAppF[F[_]: Async, Msg, Model]:
...
+        case _ =>
+          // Log a warning message?
+          ()
+    }
+  }
+
+  private def appElementFlags(tyrianAppElement: dom.HTMLElement): Map[String,String] =
+    val appFlags = for {
+      (dataAttr, attrValue) <- tyrianAppElement.dataset
+      if dataAttr.startsWith("tyrianAppFlag")
```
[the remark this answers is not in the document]
stevechy: I could change it to `tyrian-flag` or `flag` ?

tyrian/js/src/main/scala/tyrian/TyrianAppF.scala
```diff
@@ -96,3 +99,47 @@ trait TyrianAppF[F[_]: Async, Msg, Model]:
...
+      val tyrianAppElement = element.asInstanceOf[dom.HTMLElement]
+      val appSupplierOption = for {
+        appName <- tyrianAppElement.dataset.get("tyrianApp")
+        appSupplier <- appDirectory.get(appName)
+      } yield appSupplier
+      appSupplierOption match
+        case Some(appSupplier) =>
+          appSupplier().launch(tyrianAppElement, appElementFlags(tyrianAppElement))
+        case _ =>
+          // Log a warning message?
```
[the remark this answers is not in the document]
stevechy: Updated

examples/main-launcher/index.html
```diff
@@ -0,0 +1,17 @@
...
+  <meta charset="UTF-8">
+  <title>Main Launcher Example</title>
+  <script type="module" src="./target/scala-3.2.2/main-launcher-fastopt.js"></script>
+</head>
+
+<body>
+  <h1>Main launcher example</h1>
+  <div data-tyrian-app="CounterApp" data-tyrian-app-flag-initial-counter="42"></div>
+
+  <div data-tyrian-app="ChatApp" data-tyrian-app-flag-initial-message="Hello"></div>
```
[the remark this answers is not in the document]
stevechy: Removed

examples/main-launcher/src/main/scala/example/Main.scala
```diff
@@ -0,0 +1,13 @@
...
+import tyrian.Html.*
+import tyrian.*
+
+object Main {
+  def main(args: Array[String]): Unit =
+    TyrianApp.onLoad(
+      "CounterApp" -> CounterApp,
+      "ChatApp" -> ChatApp
+    )
+}
```
davesmith00000, line 13: Much better! Look at that!

tyrian/js/src/main/scala/tyrian/TyrianAppF.scala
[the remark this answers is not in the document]
davesmith00000: `tyrian-flag` seems nice?

examples/main-launcher/src/main/scala/example/CounterApp.scala
```diff
@@ -0,0 +1,46 @@
...
+
+  def view(model: Model): Html[Msg] =
+    div(
+      button(onClick(Msg.Decrement))("-"),
+      div(model.toString),
+      button(onClick(Msg.Increment))("+")
+    )
+
+  def router: Location => Msg =
+    _ => Msg.NavigateTo
```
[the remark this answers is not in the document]
davesmith00000: Here is another example taken from the `IndigoSandbox` object, just if you're interested. This one ignores internal links but follows external ones.

```scala
  def router: Location => Msg = Routing.externalOnly(Msg.NoOp, Msg.FollowLink(_))

  def update(model: Model): Msg => (Model, Cmd[Task, Msg]) =
    case Msg.NoOp =>
      (model, Cmd.None)

    case Msg.FollowLink(href) =>
      (model, Nav.loadUrl(href))
```

https://github.com/PurpleKingdomGames/tyrian/blob/main/indigo-sandbox/src/main/scala/example/IndigoSandbox.scala#L22

davesmith00000 (commented):
Looks great!

I've left more comments: It would be nice to change the data flag name, but everything else is just "if you want to." There's no real blockers here other than it occurs to me that I haven't actually tested it yet, so I should do that when I get a moment, but this is good work and I think it will be appreciated.

davesmith00000 (approved):
I've tested it - works great! Thank you very much for all the work. :pray:

PR #208, opened 2023-06-11
Fixed #205: Empty Elem + Option.toEmpty

### LINKED ISSUE #205: Provide an `empty` Elem and support rendering of `Option[Elem]`
My use case simplified is my model has an `errorMessage: Option[String]` field which I want to render when there's something there. So it would be nice to have an `empty: Elem` provided out of the box which would allow to write something like this:
```scala
errorMessage.map(span(_)).getOrElse(empty)
```
Although that `.getOrElse(empty)` can get repetitive and annoying very quickly, so some extension or implicit conversion would be great to support rendering `Option[Elem]`:
```scala
errorMessage.map(span(_))
// or
errorMessage.map(span(_)).orEmpty
```

davesmith00000:


Tvaroh (approved):
LGTM

PR #213, opened 2023-07-11
CmdHelper - Small optimisation.

diesalbla:
Replace the use of an accumulator list with a listBuffer. The advantage is that the last step to turn the `acc` listbuffer into a list can be done without allocating any objects, unlike the reverse which needs to do those allocations.

Note that this type of local use of mutable data structures, whose mutability does not scope the scope of a function body, is common in all the Scala codebase.

### DIFF
--- a/tyrian/js/src/main/scala/tyrian/runtime/CmdHelper.scala
+++ b/tyrian/js/src/main/scala/tyrian/runtime/CmdHelper.scala
@@ -3,35 +3,41 @@ package tyrian.runtime
 import cats.Applicative
 import tyrian.Cmd
 
+import scala.collection.mutable.ListBuffer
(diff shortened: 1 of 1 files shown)


davesmith00000:
Thank you @diesalbla! :smile:

I kicked off a CI build and it looks like there's a small problem, probably just a stray annotation.

diesalbla:
@davesmith00000 Could you try again?

davesmith00000 (approved):
Thanks, looks good. Indigo is full of optimisations like this but I hadn't bothered with Tyrian. Generally I prefer immutability during active development (for my sanity :wink:), but this section of code has been static for long enough that agree it makes sense to optimise it.

PR #214, opened 2023-07-11
Sub timeout - remove unused F wrappers.

diesalbla:
The external `F.delay` wrappers was not useful in any way, since they were being immediately unwrapped in the body of `task`, so they have no effect but cost allocations.

Note that it is the function itself that delays the side effectful operations on `dom.window` in acquire.

### DIFF
--- a/tyrian/js/src/main/scala/tyrian/Sub.scala
+++ b/tyrian/js/src/main/scala/tyrian/Sub.scala
@@ -175,28 +175,16 @@ object Sub:
 
   /** A subscription that produces a `msg` after a `duration`. */
   def timeout[F[_]: Sync, Msg](duration: FiniteDuration, msg: Msg, id: String): Sub[F, Msg] =
-    val acquire: F[(Either[Throwable, Msg] => Unit) => Int] =
-      Sync[F].delay { (callback: Either[Throwable, Msg] => Unit) =>
-        dom.window.setTimeout(
-          Functions.fun0(() => callback(Right(msg))),
-          duration.toMillis.toDouble
-        )
+    def task(callback: Either[Throwable, Msg] => Unit): F[Option[F[Unit]]] =
+      val handle = dom.window.setTimeout(
+        Functions.fun0(() => callback(Right(msg))),
+        duration.toMillis.toDouble
+      )
+      Sync[F].delay {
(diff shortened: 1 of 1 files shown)


tyrian/js/src/main/scala/tyrian/Sub.scala
```diff
@@ -175,28 +175,20 @@ object Sub:
...
   /** A subscription that produces a `msg` after a `duration`. */
   def timeout[F[_]: Sync, Msg](duration: FiniteDuration, msg: Msg, id: String): Sub[F, Msg] =
-    val acquire: F[(Either[Throwable, Msg] => Unit) => Int] =
-      Sync[F].delay { (callback: Either[Throwable, Msg] => Unit) =>
+    val acquire: (Either[Throwable, Msg] => Unit) => Int =
+      (callback: Either[Throwable, Msg] => Unit) =>
         dom.window.setTimeout(
           Functions.fun0(() => callback(Right(msg))),
           duration.toMillis.toDouble
         )
```
davesmith00000, line 183: I'll take another look at this in the morning when I'm a bit fresher, but my only slight concern here would be that we lose error handling. In practice, I don't think there are any errors to handle though... :thinking:

davesmith00000 (commented):
Thanks @diesalbla! I've kicked off a build, looks a lot cleaner. :+1:

tyrian/js/src/main/scala/tyrian/Sub.scala
diesalbla: So, the `Sync[F].delay` covers the creation of the function, but it does not cover an execution of the body of the function.

davesmith00000:
Good morning @diesalbla! Looks like a CI fail, how are you building Tyrian locally?

davesmith00000 (approved):
LGTM.

I had to think about that a bit. I expect it works because `handle` is eagerly evaluated and then the cancellation is delayed. It's definitely more efficient, however I'm not sure in general about the cost of the mental gymnastics of flipping mode from greedy to lazy Scala though.

In small cases like this where there are obviously (I hope!) no issues, it's probably ok. For anything much bigger I'd probably rather it was one or the other, i.e. CE or Vanilla / Lazy or Greedy.

PR #215, opened 2023-07-11
Sub - Remove use of Kleisli.

diesalbla:
In the Sub file methods, the Kleisli wrapper from cats is used as a way to compose two functions, but these can be written in a more transparent way without Kleisli.

### DIFF
--- a/tyrian/js/src/main/scala/tyrian/Sub.scala
+++ b/tyrian/js/src/main/scala/tyrian/Sub.scala
@@ -1,7 +1,6 @@
 package tyrian
 
 import cats.Functor
-import cats.data.Kleisli
 import cats.effect.kernel.Async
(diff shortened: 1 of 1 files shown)


tyrian/js/src/main/scala/tyrian/Sub.scala
```diff
@@ -84,7 +83,7 @@ object Sub:
       Observe(
         id,
         observable,
-        Kleisli(toMsg).map(f).run
+        toMsg.andThen(_.map(f))
```
davesmith00000, line 86: I'm not going to kick up a fuss over it, but I'm finding that unattractive. :sweat_smile:

Next time, maybe: `toMsg andThen { _.map(f) }`

davesmith00000 (approved):
LGTM :+1:  - thanks!

PR #219, opened 2023-07-31
File picker

hobnob:
Adds the capability to select a file by clicking a button, link, or other element in Tyrian. This works by sending a new `Cmd` (`File.select`) and waiting for the resulting `Msg`. The file can then be read with a further command of `FileReader.readText` or `FileReader.readImage`

### DIFF
--- a/tyrian/js/src/main/scala/tyrian/cmds/File.scala
+++ b/tyrian/js/src/main/scala/tyrian/cmds/File.scala
@@ -29,13 +29,13 @@ object File:
     val task = Async[F].delay {
       val input = document.createElement("input").asInstanceOf[html.Input];
       val p     = Promise[Array[dom.File]]()
+      input.setAttribute("type", "file")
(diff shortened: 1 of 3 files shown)


tyrian/js/src/main/scala/tyrian/cmds/File.scala
```diff
@@ -0,0 +1,50 @@
...
+import org.scalajs.dom
+import org.scalajs.dom.Event
+import org.scalajs.dom.document
+import org.scalajs.dom.html
+import tyrian.Cmd
+
+import scala.concurrent.Promise
+
+trait FileSelectedEvent:
+  val file: dom.File
```
davesmith00000, line 13: I don't think this is used, but not a showstopper.

davesmith00000 (approved):
LGTM, and I understand you're using it with gusto so. Nothing like a bit of dog fooding. :smile:

PR #220, opened 2023-08-04
update dependencies

ngbinh:
Some of the code changes are to make sbt-tpolcat new directives happy.


build.sbt
```diff
@@ -11,8 +11,6 @@ Global / resolvers += "Sonatype S01 OSS Snapshots" at "https://s01.oss.sonatype.
 
 ThisBuild / versionScheme := Some("early-semver")
 
-ThisBuild / scalafixDependencies += "com.github.liancheng" %% "organize-imports" % "0.6.0"
```
davesmith00000, line 14: I'd rather not lose this. It produces a warning which may now be a fatal warning. There are two fixes to that though:

1. We can convert the warnings to info level (there's a bit of sbt magic that does that but I'll have to look it up)
2. We can downgrade to version `0.5.0`.

build.sbt
```diff
@@ -22,6 +20,12 @@ lazy val scalaJsDocsVersion = "1.13.1"
 lazy val scalaDocsVersion   = "3.3.0"
 lazy val indigoDocsVersion  = "0.14.0"
 
+ThisBuild / tpolecatExcludeOptions := Set(
+  ScalacOptions.warnNonUnitStatement,
+  ScalacOptions.warnValueDiscard,
+  ScalacOptions.warnUnusedImports
+)
```
davesmith00000, line 27: Slightly unsure why we'd exclude these? Scala 3 compatibility maybe?

davesmith00000 (commented):
Hi @ngbinh, very sorry for the slow response, I've been working on other things!

Broadly looks good but I've left a couple of comments / queries.

build.sbt
davesmith00000: On testing, I've realised it's been removed because it is now built into Scalafix.

davesmith00000:
Thank you for the work. I've added a commit to remove the sbt-tpolecat excludes and upgrade to the latest version. Will check the build in the morning and hopefully merge. Sorry for the delay.

ngbinh:
@davesmith00000 great! No worry! Thanks for this great library.

PR #226, opened 2023-09-14
Fix map for Event attributes

jmcclell:
There are two issues present with the map method today:

1. Its return type is widened to Attr
2. It does not propagate all of the properties of the current Event

This means, e.g., if you want to use a pattern where sub-components use their own internal message type which then gets mapped in the parent, events do not map cleanly.

This PR fixes the issues by using the `copy` method. I have extended the usage of `copy` to the other, similar `map` methods, as well, to future-proof in case additional parameters are added to these other attribute classes as seems to have been the case with Event.


davesmith00000 (approved):
Looks great! Thank you!

PR #236, opened 2023-12-17
Fixed #232: LocationDetails shared for JVM+JS

### LINKED ISSUE #232: Consider moving `LocationDetails` into tyrian shared to publish for JVM
I'm going to make a straight up copy of it for another project, reuse would be nice.

Also `Location`.

davesmith00000:


tyrian/js/src/main/scala/tyrian/package.scala
```diff
@@ -0,0 +1,9 @@
+package tyrian
+
+extension (l: Location.type)
```
armanbilge, line 3: I'd encode this as a platform trait instead of an extension. The extension will only work if the user does `import tyrian.*`  and not if they explicitly import things.

The platform trait encoding looks like
```scala
object Location extends LocationPlatform

trait LocationPlatform: // JS
  def fromJsLocation(...): ...
  
trait LocationPlatform // JVM
  // empty

davesmith00000: Yes, that's a much better idea. Thanks!

PR #251, opened 2024-01-07
Adding API on cmd.FileReader for reading a file as ArrayBuffer

### LINKED ISSUE #249: FileReader: Missing `readAsArrayBuffer`?
Looks like `FileReader` might be missing `readAsArrayBuffer`, we have `readAsDataURL` and `readAsText` already.

https://developer.mozilla.org/en-US/docs/Web/API/FileReader/readAsArrayBuffer

KristianAN:
This commit adds implementation of support for dom.FileReader.readAsArrayBuffer as public API of the cmd.FileReader object exposing a function that reads the file as a Vector[Byte] using the typdarray api.

Updated goodies.md with new API

fixes #249

### DIFF
--- a/docs/02-guides/goodies.md
+++ b/docs/02-guides/goodies.md
@@ -7,7 +7,7 @@ Tyrian comes with a number of handy functions built-in that you can make use of 
 These nuggets of functionality are used as commands.
 
 - `Dom` - A few methods such as `focus` and `blur` to manipulate the DOM. Inspired by the Elm [Browser.Dom](https://package.elm-lang.org/packages/elm/browser/latest/Browser.Dom) package.
-- `FileReader` - Given the id of a file input field that has had a file selected, this Cmd will read either an image or text file to return an `HTMLImageElement` or `String` respectively.
+- `FileReader` - Given the id of a file input field that has had a file selected, this Cmd will read either raw bytes, an image or text file to return a `Vector[Byte]` or an `HTMLImageElement` or `String` respectively.
 - `Http` - Make HTTP requests that return their responses as a message.
 - `ImageLoader` - Given a path, this cmd will load an image and return an `HTMLImageElement` for you to make use of.
 - `LocalStorage` - Allows you to save and load to/from your browsers local storage.
--- a/docs/02-guides/goodies.md
+++ b/docs/02-guides/goodies.md
@@ -7,7 +7,7 @@ Tyrian comes with a number of handy functions built-in that you can make use of 
 These nuggets of functionality are used as commands.
 
 - `Dom` - A few methods such as `focus` and `blur` to manipulate the DOM. Inspired by the Elm [Browser.Dom](https://package.elm-lang.org/packages/elm/browser/latest/Browser.Dom) package.
-- `FileReader` - Given the id of a file input field that has had a file selected, this Cmd will read either raw bytes, an image or text file to return a `Vector[Byte]` or an `HTMLImageElement` or `String` respectively.
+- `FileReader` - Given the id of a file input field that has had a file selected, this Cmd will read either raw bytes, an image or text file to return a `IArray[Byte]` or an `HTMLImageElement` or `String` respectively.
 - `Http` - Make HTTP requests that return their responses as a message.
 - `ImageLoader` - Given a path, this cmd will load an image and return an `HTMLImageElement` for you to make use of.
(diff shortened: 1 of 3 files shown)


tyrian/js/src/main/scala/tyrian/cmds/FileReader.scala
```diff
@@ -53,16 +54,36 @@ object FileReader:
...
+    readFile(file, ReadType.AsText)(cast andThen resultToMessage)
 
-  private def readFromInputField[F[_]: Async, Msg](fileInputFieldId: String, isText: Boolean)(
+  /** Reads an input file from an input field as bytes */
+  def readBytes[F[_]: Async, Msg](inputFieldId: String)(resultToMessage: Result[Vector[Byte]] => Msg): Cmd[F, Msg] =
+    val cast: Result[js.Any] => Result[Vector[Byte]] =
+      case Result.Error(msg) => Result.Error(msg)
+      case Result.File(n, p, d) =>
+        try Result.File(n, p, new typedarray.Int8Array(d.asInstanceOf[typedarray.ArrayBuffer]).toVector)
+        catch case _ => Result.Error("File is not bytes")
```
KristianAN, line 65: Not the greatest message here, all files are bytes... but I'm not sure what else would be appropriate?

KristianAN:
Should this use the new Scala 3 IArray instead of Vector?

tyrian/js/src/main/scala/tyrian/cmds/FileReader.scala
davesmith00000: I would go with something like "Could not cast loaded file data to byte array". Specific about what went wrong but vague on the implementation details.

tyrian/js/src/main/scala/tyrian/cmds/FileReader.scala
```diff
@@ -53,16 +54,36 @@ object FileReader:
...
+
+    readFromInputField(inputFieldId, ReadType.AsArrayBuffer)(cast andThen resultToMessage)
+
+  /** Reads an input file as bytes */
+  def readBytes[F[_]: Async, Msg](file: dom.File)(resultToMessage: Result[Vector[Byte]] => Msg): Cmd[F, Msg] =
+    val cast: Result[js.Any] => Result[Vector[Byte]] =
+      case Result.Error(msg) => Result.Error(msg)
+      case Result.File(n, p, d) =>
+        try Result.File(n, p, new typedarray.Int8Array(d.asInstanceOf[typedarray.ArrayBuffer]).toVector)
+        catch case _ => Result.Error("File is not bytes")
```
davesmith00000, line 75: Same here.

docs/02-guides/goodies.md
```diff
@@ -7,7 +7,7 @@ Tyrian comes with a number of handy functions built-in that you can make use of
 These nuggets of functionality are used as commands.
 
 - `Dom` - A few methods such as `focus` and `blur` to manipulate the DOM. Inspired by the Elm [Browser.Dom](https://package.elm-lang.org/packages/elm/browser/latest/Browser.Dom) package.
-- `FileReader` - Given the id of a file input field that has had a file selected, this Cmd will read either an image or text file to return an `HTMLImageElement` or `String` respectively.
+- `FileReader` - Given the id of a file input field that has had a file selected, this Cmd will read either raw bytes, an image or text file to return a `Vector[Byte]` or an `HTMLImageElement` or `String` respectively.
```
davesmith00000, line 10: Thank you for the docs / scaladocs improvements!

davesmith00000 (approved):
My apologies for taking a while to review this.

This is excellent work! I've tested it and it works just fine. I also appreciate the small improvements you've made along the way such as the `ReadType` enum, the docs, and adding a demo to the sandbox.

To you question:

> Should this use the new Scala 3 IArray instead of Vector?

I think that's an excellent idea, yes.

I've left another minor response, but I've hit approve as I'm overall happy with the work, and it's taken me ages to do this review.

KristianAN:
Updated the error message and changed to use IArray. Note that caling toArray on an IArray produces this error message
```shell
method toArray in object IArray is deprecated since 3.0.1: This method implementation is incorrect and calling it can crash your program, please use `IArray.genericWrapArray(myIArray).toArray` instead. bloop [645, 43]
```
I don't really see this as an issue as the error message is descriptive enough.

davesmith00000 (approved):
LGTM! :tada:

Issue #201, opened 2023-05-14
Provide a sub that hooks into the main render loop


davesmith00000:
At the moment we have a Sub that provides a frame tick by setting up a stream that continuously calls `requestAnimationFrame`. Tyrian is already doing this in the background. It would be nice (for efficiency) to hook into the same loop.

PR #255, opened 2024-01-31
Make render loop wait for model updates

armanbilge:
Previously, the render loop was running on every animation frame in the browser. Now, then render loop waits until there has been at least one update to the model before re-starting the render cycle.

### DIFF
--- a/tyrian/js/src/main/scala/tyrian/runtime/ModelHolder.scala
+++ b/tyrian/js/src/main/scala/tyrian/runtime/ModelHolder.scala
@@ -1,3 +1,0 @@
-package tyrian.runtime
-
-final case class ModelHolder[Model](model: Model, updated: Boolean)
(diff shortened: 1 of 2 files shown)


davesmith00000:
Thanks @armanbilge, that is what we talked about.

I'm thinking about this in terms of https://github.com/PurpleKingdomGames/tyrian/issues/201 and wondering if this is the right strategy though. I'll try and mull it over on an up-coming dog walk or something. :smile:

davesmith00000:
@armanbilge, as it stands this PR doesn't quite work (the page never re-renders), but if you remove `resetModelUpdateCount` i.e. lines 41 and 90, then it appears to work perfectly.

I imagine what you were trying to do there is avoid a situation when you end up with an overabundance of permits available to avoid multiple renders?

davesmith00000:
image

Quite. :confused: emoji is my general feeling in that bit of code. Which is odd because it's barely 60 lines of not terribly complicated looking code. :smile: 

I'm in the middle of something at the moment, but I think when I get the chance, I'd like to see if there's some way to reformulate it that makes it easier to follow.

davesmith00000:
Closing as this issue has been resolved by another PR.

PR #256, opened 2024-02-17
Add debouncing example

DLakomy:
Recently I've been thinking how to add debouncing to an input using Tyrian. I've managed to find a good example in Elm. Perhaps something more declarative would be better, but I think this simple example is a good start. If there are any suggestions, I'll be happy to hear them.


examples/debouncing/src/main/scala/example/Main.scala
```diff
@@ -0,0 +1,67 @@
...
+    (Model("", None), Cmd.None)
+
+  def update(model: Model): Msg => (Model, Cmd[IO, Msg]) =
+    case Msg.UpdateValue(v) =>
+      (model.copy(debouncer = Some(v, DEBOUNCING_MILLIS)), Cmd.None)
+    case Msg.TimePassed =>
+      model.debouncer match
+        case Some((v, time)) =>
+          if (time < 0)
+            (model.copy(value = v), Cmd.None)
```
davesmith00000, line 28: I believe this should also reset the debouncer:
`model.copy(value = v, debouncer = None)`

examples/debouncing/src/main/scala/example/Main.scala
```diff
@@ -0,0 +1,67 @@
...
+          else
+            (model.copy(debouncer = Some((v, time - TICK_INTERVAL))), Cmd.None)
+        case None => (model, Cmd.None)
```
davesmith00000, line 31: Consider:

```scala
      model.debouncer match
        case Some((v, time)) if time < 0 =>
          (model.copy(value = v, debouncer = None), Cmd.None)

        case Some((v, time)) =>
          (model.copy(debouncer = Some((v, time - TICK_INTERVAL))), Cmd.None)

        case None =>
          (model, Cmd.None)
```

examples/debouncing/src/main/scala/example/Main.scala
```diff
@@ -0,0 +1,67 @@
...
+import tyrian.*
+
+import scala.concurrent.duration.*
+import scala.scalajs.js.annotation.*
+
+@JSExportTopLevel("TyrianApp")
+object Main extends TyrianIOApp[Msg, Model]:
+
+  val DEBOUNCING_MILLIS: Int = 500
+  val TICK_INTERVAL: Int = 100
```
davesmith00000, line 14: Stylistically in Scala it's unusual to see this sort of formatting. Sometimes people capitalise constants, so perhaps consider:

```scala
  val DebouncingMillis: Int = 500
  val TickInterval: Int     = 100
```

examples/debouncing/src/main/scala/example/Main.scala
```diff
@@ -0,0 +1,67 @@
...
+  def subscriptions(model: Model): Sub[IO, Msg] =
+    model match
+      case Model(_, Some(_)) =>
+        Sub
+          .every[IO](TICK_INTERVAL.millis, "tick")
+          .map(_ => Msg.TimePassed)
+      case _ =>
+        Sub.None
+
+final case class Model(value: String, debouncer: Option[(String, Int)])
```
davesmith00000, line 62: `(String, Int)`

Only if you want to: It would be interesting to model the debouncer more intentionally and see what it does to your code, something like (I'm making this up, untested, probably incomplete etc):

```scala
final case class Debouncer(value: String, timeout: Int, timeRemaining: Int):
  def reset: Debouncer = Debouncer("", timeout, timeout)
  def update(timePassed): Debouncer = Debouncer(value, timeout, timeRemaining - timePassed)
```

examples/debouncing/src/main/scala/example/Main.scala
```diff
@@ -0,0 +1,67 @@
...
+      (model.copy(debouncer = Some(v, DEBOUNCING_MILLIS)), Cmd.None)
+    case Msg.TimePassed =>
+      model.debouncer match
+        case Some((v, time)) =>
+          if (time < 0)
+            (model.copy(value = v), Cmd.None)
+          else
+            (model.copy(debouncer = Some((v, time - TICK_INTERVAL))), Cmd.None)
+        case None => (model, Cmd.None)
+    case _ => (model, Cmd.None)
```
davesmith00000, line 32: Strongly preferred not to default, especially when you only have one clause left and it's a NoOp, because we want to encourage people to use exhaustive matching on `Msg` types.

```scala
case Msg.NoOp => (model, Cmd.None)
```

davesmith00000 (approved):
This is great, I love it! :tada: 

I have approved because it's good enough as it is, and I'm delighted to have the contribution. I have however, also left some comments for your consideration. Mostly style points to encourage people to use best practices. The only comment that might be a technical issue, is that I think you need to reset the debouncer once you release the value.

examples/debouncing/src/main/scala/example/Main.scala
DLakomy: True, I've forgotten about it. Fixed.

examples/debouncing/src/main/scala/example/Main.scala
DLakomy: I agree that it's more readable. Fixed.

examples/debouncing/src/main/scala/example/Main.scala
DLakomy: Done. And this time I remembered to run the formatter

examples/debouncing/src/main/scala/example/Main.scala
DLakomy:
```scala
final class Debouncer[A](val value: A, timeRemaining: Int):
  def update(timePassed: Int): Debouncer[A] =
    Debouncer(value, timeRemaining - timePassed)
  def isSteady = timeRemaining <= 0
```
I've implemented something like this (+ necessary changes in other parts of code; the model now has `debouncer: Option[Debouncer[String]]`). I haven't pushed this changes, because to be honest I'm not sure if it's better than the original example (in terms of being easy to follow). Some logic is better encapsulated, but `subscriptions` and `update` definitions still need to be aware of the `Debouncer` and actively manage it's state.

Perhaps something like `onInput(Msg.UpdateValue(_)).debounced(100.millis)`, with no additional logic in `subscriptions` and `update`, would be ideal (if I remember correctly, in Laminar the API is similar to this). Before writing the example I tried to come up with something like this (possibly involving some changes to Tyrian itself), but for now I haven't had an idea nor enough time :( Hence I decided to at least provide an example, as a proof of concept.

davesmith00000: > but subscriptions and update definitions still need to be aware of the Debouncer and actively manage it's state.

Yes, that's expected, it's part of how the Elm architecture scales that, if you want to use this pattern, you just need to accept it and go along with it. However you construct a "component" like this, you always have to manually tie it into the various systems that will update it, somehow. But as I say, that's ok, it's just part of the pattern and you get used to it.

But I also think not including it is completely fine. If this version is easier to follow, that's totally ok with me. Thanks for giving it a shot! :+1:

davesmith00000 (approved):
LGTM!

Issue #254, opened 2024-01-30
Investigate: Tyrian apps seem CPU intensive


davesmith00000:
If you run the standard g8 template, and then check the task manager in Chrome, you can see the Tyrian is using a surprising amount of CPU even when idling.

If may by an unavoidable overhead of the architecture, but maybe we just introduced a regression somewhere. Worth looking into.

davesmith00000:
We speculate that this is due to the hot render loop.

https://github.com/PurpleKingdomGames/tyrian/pull/255

davesmith00000:
I've spent some time butchering the runtime and trying out different things.

My own linux machine seems to show even higher CPU load than I recall the original reporter talking about.

Some findings:

1. If I remove the render loop, CPU usage when the app is at rest, is zero as expected. So that's good.
2. If I remove the normal render loop, and bolt in a "does nothing" loop based on pure Scala.js (no Cats Effect, does not render with Snabbdom), then the CPU load is back.

So fundamentally, the CPU load seems to be coming from `requestAnimationFrame`. Below you can see the "does nothing" version in the screenshot, note the CPU load. 

![image](https://github.com/PurpleKingdomGames/tyrian/assets/908709/fe37a7de-94aa-4a87-aaa9-5bc5c8dfd169)

Surely this can't be the case?! How does Elm cope? Well on my machine... Elm has almost the same problem. Maybe slightly better, depending on the type of app your running, but not much. Certainly when I'm interacting with an app there is no real difference

Still, is there anything we can do?

Well, we render on frame tick if there has been a model change. We do that because if you're doing heavy updates you don't want to be re-rendering for nothing, but it adds constant load. If you render only on `Msg` then you can have zero overhead while the app is idle, but it can go nuts during use - for instance it would render even on a No-Op.

The only other thing I can think of is to do a hybrid. In this model you render on requestAnimationFrame, as long as you saw a message in the last, say, ~2 seconds. So if your app is idle, the rendering will go idle, but while stuff is happening (whether triggered by the user or anything else) the rendering continues.

armanbilge:
> The only other thing I can think of is to do a hybrid. In this model you render on requestAnimationFrame, as long as you saw a message in the last

That is what my PR https://github.com/PurpleKingdomGames/tyrian/pull/255 attempted to do. I got side-tracked but I intend to take a second look and fix it :)

davesmith00000:
No worries @armanbilge.

[...]

PR #263, opened 2024-03-28
Reworking the runtime

davesmith00000:
This PR may not survive, I'm just using it so it's easier for me to check my own work.

So far I've done some refactoring. No suggestion that this is an improvement, but the exercise has helped me reacquaint myself with what's going on.

The idea is to do the following, from the [issue comment](https://github.com/PurpleKingdomGames/tyrian/issues/254#issuecomment-2019863143):

> In this model you render on requestAnimationFrame, as long as you saw a message in the last, say, ~2 seconds. So if your app is idle, the rendering will go idle, but while stuff is happening (whether triggered by the user or anything else) the rendering continues.

Rough plan of attack:

Step 1 - Update pokes the renderer:

- Remove ModelHolder
- A model update triggers redraw
- Redraw is on a Renderer class and returns a new instance (to be stored in a Ref?)
- Render has a finite state machine: Idle, Running
- When a redraw happens you record a timestamp, update the state to running (always, for now), then immediately render

Step 2 - Render runs on a loop (again):

- Instead of the update poking the renderer and a render being triggered, it instead checks:  if running do nothing, if not running, start render loop and set state to running
- Render loop checks if < 2 seconds since last update, if true, loop, else state to idle and don't loop.

Something like that.

### DIFF
--- a/tyrian/js/src/main/scala/tyrian/runtime/ModelHolder.scala
+++ b/tyrian/js/src/main/scala/tyrian/runtime/ModelHolder.scala
@@ -1,3 +1,0 @@
-package tyrian.runtime
-
-final case class ModelHolder[Model](model: Model, updated: Boolean)
--- a/tyrian/js/src/main/scala/tyrian/runtime/Renderer.scala
+++ b/tyrian/js/src/main/scala/tyrian/runtime/Renderer.scala
@@ -1,0 +1,89 @@
+package tyrian.runtime
+
+import cats.effect.kernel.Async
+import cats.effect.kernel.Clock
+import cats.effect.kernel.Ref
+import cats.effect.std.Dispatcher
+import cats.syntax.all.*
+import org.scalajs.dom
+import snabbdom.VNode
+import tyrian.Html
+import tyrian.Location
+
+final case class Renderer(vnode: VNode, state: RendererState):
+
+  def runningAt(t: Long): Renderer =
+    this.copy(state = RendererState.Running(t))
+
+object Renderer:
+
+  def init[F[_]](vnode: VNode)(using F: Async[F]): F[Ref[F, Renderer]] =
+    F.ref(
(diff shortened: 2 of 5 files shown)


davesmith00000:
After going round and round trying to do this with CE, I decided just to PoC it imperatively to see if it was going to work at all. It kinda does.

My CPU has always seemed a little more eager that others have reported, so I wouldn't worry that the number seems to jump quite high in the gif, or that it's a little twitchy. This is similar behaviour to what I see on this machine from Elm apps.

Will need to verify on other machines.

tyrian_reduced_rendering_cpu

davesmith00000:
Here you can see where the frame animations kick in when activity occurs, and drops off when the user stops doing work.

image

davesmith00000:
Screenshot from the Cats Effect version, which has replaced the imperative implementation.

image

davesmith00000:
If you have any free time to take a look at this @armanbilge, that would be great (don't worry if not :smile:). This appears to do the trick but probably doesn't do it the nicest way.

davesmith00000:
From my testing, this appears to be a vast improvement on pages that are idle between user interactions. I can't tell if there's any real impact on pages that do constant work - like games.

We don't have a way of defining "options" for Tyrian yet (like Indigo's `GameConfig`), but I'm wondering if it would be beneficial to be able to switch modes when you _know_ what kind of site you've got. The "low consumption when idle" mode developed here could be a default.

hobnob:
This looks like it'll work in a vast majority of cases - as you say, I'm not sure about games, where the model might be updated a fair bit. That said - I think we should put it live as is, as I suspect most people will be using Tyrian for static sites, and Indigo for games. In my experience, even if you use Indigo for games and Tyrian as the UI, the actual amount of time you spend updating Tyrian's model is far far less than Indigo

hobnob (approved):

Issue #266, opened 2024-05-05
Unable to upload an image file to the server via HTTP


a-khakimov:
I couldn't figure out how to upload an image file to the server using the HTTP method with Tyrian. I couldn't find any code examples. Help me please to understand how to do this correctly.

Questions:

- How can I correctly upload an image file to the server using the Tyrian?
- Are there any limitations or peculiarities with Tyrian when transmitting files? (Currently, I'm concerned because I only saw the possibility of transmitting the following types of request body in the source code: html, json, xml, plainText)
- If the functionality is not supported, are there plans to add support for file transmission in future versions of Tyrian?

Thank you for your help!

davesmith00000:
Hi @a-khakimov, sorry for the slow reply and thank you for the pull request, I shall take a look!

The short answer is that I don't know off the top of my head. It is quite probably an oversight and that you are the first person to try this with Tyrian's HTTP client. If that's the case we should of course fix it, thank you for reporting. :pray:

I feel like @hobnob must have done stuff like this with Tyrian though? Maybe we can summon him by invoking his name... :smile: 

I don't feel great about having to recommend this, just because we should support images out of the box, but as a work-around you can also use  http4s-dom (or any other Cats Effect / ZIO compatible client), which is a more fully featured HTTP Client. There is an example here: https://github.com/PurpleKingdomGames/tyrian/tree/main/examples/http4s-dom

davesmith00000:
I'm looking over your PR and reading up on the subject. There are actually two workarounds at the moment, both with pros and cons:

1. As above, use http4s-dom
2. Base64 encode the image and send it as plaintext.

PR #267, opened 2024-05-05
Http request with BodyInit

a-khakimov:
I made a small draft that solves my problem #266. I suggest making such a change to support http requests with a body for other types, not only text.

### DIFF
--- a/tyrian/src/main/scala/tyrian/http/Models.scala
+++ b/tyrian/src/main/scala/tyrian/http/Models.scala
@@ -1,5 +1,7 @@
 package tyrian.http
 
+import org.scalajs.dom.BodyInit
+
 /** An Error will be returned if something goes wrong with an HTTP request. */
 enum HttpError:
   /** A BadRequest means that the provide request was not valid for some reason.
@@ -41,6 +43,7 @@ enum Body derives CanEqual:
     *   the content of the body
     */
   case PlainText(contentType: String, body: String) extends Body
+  case File(contentType: String, body: BodyInit) extends Body
 
 object Body:
   def html(body: String): Body      = Body.PlainText("text/html", body)
--- a/tyrian/src/main/scala/tyrian/http/Http.scala
+++ b/tyrian/src/main/scala/tyrian/http/Http.scala
@@ -51,6 +51,9 @@ object Http:
         case (Body.PlainText(contentType, body), method) if !Set(Method.Get, Method.Head).contains(method) =>
           headers.append("Content-Type", contentType)
           requestInit.body = body
+        case (Body.File(contentType, body), method) if !Set(Method.Get, Method.Head).contains(method) =>
+          headers.append("Content-Type", contentType)
+          requestInit.body = body
         case _ =>
 
       requestInit.headers = headers


davesmith00000:
Hi @a-khakimov - sorry for the slow reply. I've been stuck on a long piece of work in another repo. Will check it out as soon as I get back to Tyrian. :+1:

a-khakimov:
Hi @davesmith00000! Just wanted to check in regarding PR and see if you’ve had a chance to take a look. Thanks, and appreciate your time!

tyrian/src/main/scala/tyrian/http/Models.scala
```diff
@@ -41,6 +43,7 @@ enum Body derives CanEqual:
     *   the content of the body
     */
   case PlainText(contentType: String, body: String) extends Body
+  case File(contentType: String, body: BodyInit) extends Body
```
davesmith00000, line 46: This might be ok, but I'm a bit worried about using BodyInit as a catch-all type. I am entirely open to counter arguments. 🙂 

I haven't tried this so this is speculation, but from reading around I think we need something like this (though this seems cumbersome):

```scala
enum Body derives CanEqual:
  case Empty
  case PlainText(contentType: String, body: String)
  case File(contentType: String, body: org.scalajs.dom.File)
  case Blob(contentType: String, body: org.scalajs.dom.Blob)
  case FormData(contentType: String, body: org.scalajs.dom.FormData)
  case ArrayBuffer(contentType: String, body: org.scalajs.dom.ArrayBuffer)
  case TypedArray(contentType: String, body: org.scalajs.dom.TypedArray)
  case URLSearchParams(contentType: String, body: org.scalajs.dom.URLSearchParams)
  case ReadableStream(contentType: String, body: org.scalajs.dom.ReadableStream)
```

It also feels like we could do with an enum of content types, and then we could provide helper methods so that you could do things like:

```scala
def bodyImagePng(data: Blob): Body =
  Body.Blob(ContentType.imagePng, data)
```

Or something. What do you think?

hobnob: A list of content types is a [big 'ol list](https://developer.mozilla.org/en-US/docs/Web/HTTP/MIME_types/Common_types) - maybe just types we support? I suspect for now you probably just want `PlainText`, `File`, and `Blob` to start.. dealing with `application/json` and `application/xml` are problems all on their own for another time :slightly_smiling_face:

davesmith00000: Possibly crossed wires (or I'm just wrong, wouldn't be the first time 😁 ), we're not talking mime-types, we're talking bodyinit types - I think that's the whole list ☝️.

References:
https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/send#parameters
https://fetch.spec.whatwg.org/#bodyinit-unions

> An XMLHttpRequestBodyInit, which per the Fetch spec can be a Blob, an ArrayBuffer, a TypedArray, a DataView, a FormData, a URLSearchParams, or a string.

Not...sure...where I got `ReadableStream` from... 🤔 

Does that change anything for you?

davesmith00000 (approved):
I'm approving this because I've been unintentionally sitting on it. It is an improvement that solves a problem, thank you for the work and I'm very sorry it's taken such a long time to approve.

I still think there's more to do here, but I've left a placeholder issue that points back to here for some future person to look at:
https://github.com/PurpleKingdomGames/tyrian/issues/290

I need to figure out how to rerun this build to make sure it's passing before I merge.

Thanks again!

Update: Can't figure out how to re-run it but I've run it locally and it seems ok.

PR #279, opened 2024-08-17
Refactor FileReader to fix side-effect and provide extra context

sloshy:
Found this unexpected behavior when using the FileReader by defining a `Cmd` as a Scala `val`.

The code to read the `dom.File` from an input element by ID executes eagerly, before returning the final suspended `Cmd` for reading a file. This means that if you use this helper in your code as a `val` it will get the result at definition time, and not when you actually expect the `Cmd` to execute.

To fix this, I made the following changes:
* `readFile` now takes a parameter to get the `dom.File` input so that it can be provided multiple ways.
* There is a new `Result.NoFile` type that is a variant of `Result.Error` but providing the additional context that there is no file specified. This can be useful for more precise application error messages in addition to making refactoring easier.
* I extracted the `cast` functions in each helper method to their own values so the file size is smaller and it's easier to keep these in sync if their behavior should ever change.

Hope this change meets your standards and thanks again for the wonderful project!

### DIFF
--- a/tyrian/src/main/scala/tyrian/cmds/FileReader.scala
+++ b/tyrian/src/main/scala/tyrian/cmds/FileReader.scala
@@ -1,6 +1,7 @@
 package tyrian.cmds
 
 import cats.effect.kernel.Async
+import cats.syntax.all.*
 import org.scalajs.dom
 import org.scalajs.dom.Event
 import org.scalajs.dom.document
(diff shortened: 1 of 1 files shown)


davesmith00000:
Looks great, thanks @sloshy! I'm going to read it properly tomorrow when I'm slightly more awake, but I've kicked the build off in the meantime. :smiley:

tyrian/src/main/scala/tyrian/cmds/FileReader.scala
```diff
@@ -16,103 +18,88 @@ import scala.scalajs.js.typedarray
   */
 object FileReader:
 
+  val readImageCast: Result[js.Any] => Result[String] =
```
kubukoz, line 21: question: should these `*Cast` vals be private?

tyrian/src/main/scala/tyrian/cmds/FileReader.scala
```diff
@@ -16,103 +18,88 @@ import scala.scalajs.js.typedarray
...
-        case ReadType.AsText => fileReader.readAsText(file)
-        case ReadType.AsArrayBuffer =>
-          fileReader.readAsArrayBuffer(file)
-        case ReadType.AsDataUrl => fileReader.readAsDataURL(file)
-
-      p.future
+    val task = maybeGetFile.flatMap {
+      case None => Future.successful(Result.NoFile("No files on specified input")).pure[F]
+      case Some(file) =>
+        Async[F].delay {
```
kubukoz, line 75: thought: the sequence of `callback -> promise -> future -> f` seems like it should be `Async[F].async`. On top of the conversions having extra cost, it doesn't handle cancellation (which might result in memory leaking the event listener/promise in case the command gets cancelled).

It would probably be beneficial to rewrite this as `Async[F].async` and remove the event listener in case of interruption. note, I can't remember if command effects actually are interruptible.

tyrian/src/main/scala/tyrian/cmds/FileReader.scala
davesmith00000: I like this improvement of extracting them out in general, but yes I agree that they don't belong on the public interface. :+1:

tyrian/src/main/scala/tyrian/cmds/FileReader.scala
davesmith00000: > I can't remember if command effects actually are interruptible

In practice, not really. You can do it but you have to work hard to make it happen, it isn't just easily available. It is one of those topics that comes up a lot on the backend but... my hypothesis is that it's less important on the frontend, and that if you have one of those cases where it really does matter then you won't mind doing the work.

So it's a tradeoff. In this arch style, precise effect management is more difficult, but the overall code and life cycle is (IMO) simpler and cleaner (Discussions on scaling not withstanding :stuck_out_tongue:).

davesmith00000: I agree that there is a good argument for converting to `Async[F].async`, however I don't think that's @sloshy's problem. This code is an improvement, so no need to hold up the PR any longer.

But it's a good suggest, and I've [captured it as an issue](https://github.com/PurpleKingdomGames/tyrian/issues/280).

davesmith00000 (approved):
LGTM @sloshy, thanks for the work. Will hang on before merging in case anyone wants to comment. If you feel like applying the suggesting to make the extracted functions private, that would be lovely, but otherwise I'm happy. Thanks again!

tyrian/src/main/scala/tyrian/cmds/FileReader.scala
sloshy: Have been rather busy recently but I'll make a quick commit for this later tonight (or try to)

davesmith00000: Thanks @sloshy, I've been busy too, it happens. :slightly_smiling_face: 

Don't worry about it. I'm going to go ahead and merge this and then I'll quickly make the change - your efforts are appreciated! :pray:

sloshy: Ahh sorry things have been a little hectic over here. You're the best!

PR #284, opened 2024-08-26
Support scala-js-snabbdom-0.2.0-M3 and lazy property setting

sloshy:
This PR comes from troubleshooting strange behavior in a webapp I'm making where the value property on some elements could sometimes not appear to update as I'd expect. In trying multiple solutions such as setting keys on the elements I'm updating, the only solution I've found that works is a combination of the following:

* updating scala-js-snabbdom to 0.2.0-M3
* Adding a way to hint to tyrian that you want properties applied later in the rendering process (`Html#setLazy`)

Part of the reason why this works is that the behavior of the `Props` module in scala-js-snabbdom [has changed between versions](https://github.com/buntec/scala-js-snabbdom/compare/v0.1.0..v0.2.0-M3#diff-f1c471db1a0dcc4fe670fdde143206e9be24f72721febc79b7eb28407ae71dc6). In particular it no longer filters out the `value` property on updates. Trying to add `setLazy` to the branch supporting v0.1.0 does not fix the issue.

The way this is meant to work is, you will add `.setLazy` onto any elements that are exhibiting strange property-setting behavior, to delay setting those properties until after every other element is updated. If it's added to a normal element, it may change behavior, so it's only meant to be used in strange edge-cases where values or other properties appear as if they are not correctly being set.

Alternatively -- if you can think of a more elegant way to solve the issues I've been experiencing without this kind of hack-ish solution, or without supporting a milestone-version of this library, I would love to hear it. I'm personally a bit nervous about this requiring a milestone version of scala-js-snabbdom and it has not been updated in quite some time. That being said, the behavior has not actually changed all that much from v0.1.0 from the looks of things and everything else appears to work as I'd expect so far, so YMMV.

### DIFF
--- a/docs/02-guides/html.md
+++ b/docs/02-guides/html.md
@@ -145,3 +145,23 @@ One occasion where you may need to make your own tags or attributes etc., is whe
 A previous case of this was where the input field `value` `property` was incorrectly declared as an `attribute`, and so the value wasn't changed as expected based on model updates.
 
 Effort has gone in to getting these things right, but if you come across any issues, [please report them](https://github.com/PurpleKingdomGames/tyrian/issues), and the workaround is to re-declare it yourself as above while a fix is produced.
+
+#### Known Gotcha: Eager Property Setting
+
+It's possible that sometimes properties of elements will appear inconsistently applied due to how changes are applied to the DOM.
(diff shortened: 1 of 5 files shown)


sloshy:
After digging a bit more into how I'm supposed to use keys, it appears this is entirely unnecessary. The issue I was experiencing was solved by making sure that all of the children of the select element also had keys.

In my defense I've been a bit sick recently so maybe my head's messed up. No worries and thanks again for the great project!

Issue #293, opened 2025-01-03
String tag implicit Conversion


linux-root:
```scala
  div(
        id := "faq-tab",
        attr("data-tabs-target") := "#faq",
)(button(id:="btn-id", cls:="btn")() ,  "FAQ")
```
It appears that the string tag "FAQ" in this case is not implicitly converted to tyrian.Elem[M]; we need to explicitly define it as text("FAQ"). I’m not sure if this is the issue or if there is a reason behind this design. Should we consider improving the syntax?

davesmith00000:
Hi @linux-root,

Happy to answer questions here, but you might also try our discord server: https://discord.gg/b5CD47g

Back to the issue, if I rewrite it like this:

```scala
div(
  id := "faq-tab",
  attr("data-tabs-target") := "#faq",
)(
  List(
    button(id:="btn-id", cls:="btn")(),
    "FAQ"
  )
)
```

We can see that what's happening is that it's trying to make a list of `Elem[?]` but it's ending up with a list of type `Any`. IIRC, the syntax allows for a single element to be a `String` and that will be handled, but I can't do that with a heterogeneous arbitrary list.

To make that work, we'd need to include implicit conversions in the syntax import like this (... err... it's been a long time, don't know the Scala 3 syntax, here's Scala 2):

```scala
implicit def strToElm[Msg](str: String): Elem[Msg] = text(str)
```

Or something (not sure about that type param!). Couple of things there:

1. That sort of thing is discouraged in Scala 3, I think you have to explicit know to import implicit conversions - not sure need to check.
2. The reason its discouraged is that it's generally discouraged as a practice (although I recently heard that Mill does this a lot to smooth out the API...). We could bring in extension methods so you could do `"FAQ".text`... but is that any better? You still have to know that that is an option / available.

linux-root:
Hi @davesmith00000,
I’ve created this  [PR-294]( https://github.com/PurpleKingdomGames/tyrian/pull/294) for discussion purpose. While implicit conversions are sometimes discouraged, they are not inherently bad in all cases. In Scala 3, we have a new syntax for defining implicit conversions, which requires careful use to avoid creating confusing problems. However, in this specific case, I believe it’s perfectly appropriate.

[...]

PR #294, opened 2025-01-05
literal  tag implicit Conversion

linux-root:


tyrian-tags/shared/src/main/scala/tyrian/Html.scala
```diff
@@ -40,6 +40,10 @@ sealed trait Html[+M] extends Elem[M]:
   */
 object Html extends HtmlTags with HtmlAttributes with AttributeSyntax:
 
+  given Conversion[String, Elem[Nothing]] = text(_)
+  given Conversion[Int, Elem[Nothing]]    = n => text(n.toString)
+  given Conversion[Double, Elem[Nothing]] = d => text(d.toString)
```
davesmith00000, line 45: Two things:

1. I think this code needs to move to the `syntax` object.

https://github.com/PurpleKingdomGames/tyrian/blob/main/tyrian/src/main/scala/tyrian/syntax.scala

Then, I'm hoping, the import will be more guessable as `import tyrian.syntax.given`

2. It might be unavoidable or a non-issue, but I would be nice to keep the `Nothing` out of there. Does something like this work?

`given [Msg]: Conversion[String, Elem[Msg]] = text(_)`

I can see how it might not, but worth a try.

tyrian-tags/shared/src/test/scala/tyrian/HTMLRenderingTests.scala
```diff
@@ -145,4 +145,31 @@ class HTMLRenderingTests extends munit.FunSuite {
...
+    )(
+      "Messages",
+      span(
+        cls := "inline-flex items-center"
+      )(2)
+    )
+
+    println(example1.render)
+    println(example2.render)
+  }
```
davesmith00000, line 173: I think we'll need a new `syntaxTests` suite (assuming you move the code as suggested) and the tests will need to assert something.

davesmith00000 (commented):
Looks promising, seems worth polishing up. 😄

davesmith00000:
This approach is probably worth trying (avoid implicit conversions), see my comment underneath it, though:

https://github.com/PurpleKingdomGames/tyrian/issues/293#issuecomment-2571597718

tyrian-tags/shared/src/main/scala/tyrian/Html.scala
linux-root: 1. I've  moved code to https://github.com/PurpleKingdomGames/tyrian/blob/main/tyrian-tags/js/src/main/scala/tyrian/syntax.scala. It seems like a more appropriate place but I'm not sure. Please advice 😅 
2. I’ve changed the type from Elem[Nothing] to Text, which is the exact target of the conversion. The type `Text` won’t break type inference (e.g., resulting in an Any type) regardless of what the Msg type is. Therefore, the type parameter `[Msg]` is unnecessary in this context
```scala
  given Conversion[String, Text] = text(_)
  given Conversion[Int, Text]    = n => text(n.toString)
  given Conversion[Double, Text] = d => text(d.toString)
```

tyrian-tags/shared/src/test/scala/tyrian/HTMLRenderingTests.scala
linux-root: @davesmith00000 I completely agree. Ultimately, a test must assert something. I kept that in mind while creating this.

davesmith00000 (approved):

PR #310, opened 2025-03-06
Add OnMouseEnter and OnMouseLeave events

clemniem:
Referrences to the events:
https://developer.mozilla.org/en-US/docs/Web/API/Element/mouseenter_event

https://developer.mozilla.org/en-US/docs/Web/API/Element/mouseleave_event


clemniem:
In my local project I created: 
```scala
  def onMouseLeave[M](msg: Tyrian.MouseEvent => M): Event[Tyrian.MouseEvent, M] =
    AttributeSyntax.onEvent("mouseleave", msg)

  def onMouseEnter[M](msg: Tyrian.MouseEvent => M): Event[Tyrian.MouseEvent, M] =
    AttributeSyntax.onEvent("mouseenter", msg)
```

And it works.

I did not see if you have specs for the Attributes.

davesmith00000 (approved):
Thanks @clemniem! 😄

